PR #105, opened 2026-07-20
feat(core): collapse recursive frames — StackOverflowError reports name the recursion

### LINKED ISSUE #101: feat(core): collapse recursive frames — a StackOverflowError report is 15 identical lines
A `StackOverflowError` produces the single worst log output in Java — a thousand identical
frames. It is exactly the case where distillation should shine, and today it doesn't.

## The problem

For a recursion in *your own* code, the report currently looks like this:

```
stack (distilled, 15 of 1024 frames):
  PricingService.discountFor(PricingService.java:51) ← culprit
  PricingService.discountFor(PricingService.java:51)
  PricingService.discountFor(PricingService.java:51)
  … (12 more identical lines)
  … 1009 more frames
```

It is shorter than the raw trace, but it is *dumb*: it repeats one line 15 times and never
says the word **recursion**, which is the only thing that matters in this error.

## Why it happens

`StackDistiller.renderFrames` only collapses runs of **framework** frames. See
[`StackDistiller.java`](../blob/main/stacktale-core/src/main/java/io/github/gabrielbbaldez/stacktale/StackDistiller.java):

- line ~108: `boolean mustShow = i == 0 || i == culpritIdx || isAppFrame(el);`
- line ~120: the collapsing loop runs `while (... && !isAppFrame(frames[i]))`

In a `StackOverflowError` the premise inverts — the noise *is* your own code. Every frame is
an app frame, so `mustShow` is true for all of them and the collapsing branch never runs.
The distiller protects you from other people's noise, not from your own.

## Desired output

Simple self-recursion:

```
stack (distilled, 2 of 1024 frames):
  PricingService.discountFor(PricingService.java:51) ← culprit
  … recursion ×1023 (PricingService.discountFor)
```

Mutual recursion (`a → b → a → b …`) should name the whole cycle, because the cycle *is* the
diagnosis:

```
  … recursion ×341 (OrderService.confirm → PricingService.apply → OrderService.confirm)
```

## Suggested approach

1. In `renderFrames`, before the existing framework-collapsing logic, detect a **repeating
   run**: walk forward while the frame at `i` equals the frame at `i - period` (compare
   class + method + line), trying small periods first (1, then 2, 3, … up to ~8). A period
   of 1 is self-recursion; a larger period is a cycle.
2. Only emit the marker when the run is long enough to be meaningful (suggest ≥ 3 repeats)
   — otherwise leave today's behaviour untouched.
3. Emit `… recursion ×N (<cycle>)` where `<cycle>` lists the distinct methods of one period,
   joined by ` → `.
4. Keep the culprit frame itself visible above the marker (as in the examples).
5. The recursion check should apply to app frames *and* framework frames — a proxy or
   serializer looping is the same bug shape.

## Tests

Add to `StackDistillerTest` (and/or `StackDistillerEdgeTest`) in `stacktale-core`:

- self-recursion: build a `StackTraceElement[]` with the same element repeated 1000× and
  assert one `recursion ×` line, not 15 duplicates;
- mutual recursion: alternate two elements and assert the arrow-joined cycle;
- **no false positives**: a normal stack where a method legitimately appears twice
  (e.g. a recursive tree walk of depth 2, or a retry) must NOT be collapsed — this is the
  assertion that matters most;
- a real `StackOverflowError` caught from an actually-recursive test method.

You can build the frames by hand — `new StackTraceElement("com.acme.Svc", "loop", "Svc.java", 51)`
— and set them with `throwable.setStackTrace(frames)`. Several existing tests already do this;
`StacktaleAppenderIntegrationTest#stormControlBoundsAFloodOfDistinctErrors` is an example.

## Gotcha

The `st/1` output is pinned by golden files in `stacktale-core/src/test/resources/golden/`.
This change is **additive** (a new marker line shape) and should not alter any existing
golden file — if one changes, that is a signal the detection is firing when it shouldn't.
Run `mvn -pl stacktale-core test` and check the golden tests specifically.

Reproduce locally with a plain recursive method, and read the resulting `errors-ai.log`.

dchaudhari7177:
## What

Implements #101. `renderFrames` only collapsed runs of *framework* frames; in a `StackOverflowError` the noise is the app's own code, so every frame was `mustShow` and the report printed 15 identical lines without ever saying **recursion**.

## How

Before the existing framework-collapsing logic, detect a repeating run at the current index: walk forward while `frames[i]` equals `frames[i - period]` (class + method + line), trying periods 1..8 smallest-first — exactly the suggested approach. Then:

- **≥ 3 full cycles required** to collapse; anything shorter renders exactly as today (the no-false-positive rule)
- one full cycle stays visible above the marker, culprit annotation included
- self-recursion: `… recursion ×1023 (PricingService.discountFor)`
- mutual recursion names the whole cycle: `… recursion ×98 (OrderService.confirm → PricingService.apply → OrderService.confirm)`
- the check runs on app *and* framework frames (a looping proxy/serializer collapses the same way)
- if the culprit sits inside the run beyond the first cycle, the run is cut at the culprit so it stays visible

## Tests

Four new tests in `StackDistillerTest`:
- 1000× identical frame → culprit line + one `… recursion ×999` marker, and the frame's location appears exactly once
- alternating pair → arrow-joined cycle label, `×98`
- **no false positives**: depth-2 tree walk + a retry pair + main → 5 lines, no `recursion` marker
- a real `StackOverflowError` caught from an actually-recursive test method → one marker naming `StackDistillerTest.recurse`, report ≤ 6 lines

## Golden files

`mvn -pl stacktale-core test` passes (all suites, golden tests included) with **zero changes** under `src/test/resources/golden/` — the detection never fires on the existing corpus, per the issue's gotcha.

Fixes #101

### DIFF
--- a/stacktale-core/src/main/java/io/github/gabrielbbaldez/stacktale/StackDistiller.java
+++ b/stacktale-core/src/main/java/io/github/gabrielbbaldez/stacktale/StackDistiller.java
@@ -107,14 +107,22 @@ final class StackDistiller {
             // Recursion first: a StackOverflowError's noise is the app's own
             // code, which the framework collapsing below never touches. Keep
             // one full cycle visible (including the culprit), then mark the rest.
-            RecursionRun recursion = detectRecursion(frames, i, culpritIdx > i ? culpritIdx : frames.length);
+            // Past the frame cap the tail marker below handles everything,
+            // including further recursive runs — no trailing marker pile-up.
+            RecursionRun recursion = shown < MAX_SHOWN_FRAMES
+                    ? detectRecursion(frames, i, culpritIdx > i ? culpritIdx : frames.length)
+                    : null;
             if (recursion != null) {
                 for (int k = 0; k < recursion.period && shown < MAX_SHOWN_FRAMES; k++) {
                     out.add(location(frames[i + k]) + (i + k == culpritIdx ? " ← culprit" : ""));
                     shown++;
                 }
-                out.add("… recursion ×" + (recursion.frameCount - recursion.period)
-                        + " (" + cycleLabel(frames, i, recursion.period) + ")");
+                // Frames, not cycles: "98 recursive frames" of a 2-frame cycle
+                // is 49 repetitions — counting frames keeps the unit identical
+                // to the neighbouring "… N collapsed (…)" marker and avoids
+                // rounding when a run isn't a whole multiple of the period.
+                out.add("… " + (recursion.frameCount - recursion.period)
+                        + " recursive frames (" + cycleLabel(frames, i, recursion.period) + ")");
                 i += recursion.frameCount;
                 continue;
             }
--- a/stacktale-core/src/main/java/io/github/gabrielbbaldez/stacktale/StackDistiller.java
+++ b/stacktale-core/src/main/java/io/github/gabrielbbaldez/stacktale/StackDistiller.java
@@ -104,6 +104,20 @@ final class StackDistiller {
         int shown = 0;
         int i = 0;
         while (i < frames.length) {
+            // Recursion first: a StackOverflowError's noise is the app's own
+            // code, which the framework collapsing below never touches. Keep
+            // one full cycle visible (including the culprit), then mark the rest.
+            RecursionRun recursion = detectRecursion(frames, i, culpritIdx > i ? culpritIdx : frames.length);
+            if (recursion != null) {
+                for (int k = 0; k < recursion.period && shown < MAX_SHOWN_FRAMES; k++) {
+                    out.add(location(frames[i + k]) + (i + k == culpritIdx ? " ← culprit" : ""));
+                    shown++;
+                }
+                out.add("… recursion ×" + (recursion.frameCount - recursion.period)
+                        + " (" + cycleLabel(frames, i, recursion.period) + ")");
+                i += recursion.frameCount;
+                continue;
+            }
             StackTraceElement el = frames[i];
             boolean mustShow = i == 0 || i == culpritIdx || isAppFrame(el);
             if (mustShow) {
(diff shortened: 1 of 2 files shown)


GabrielBBaldez (commented):
Really nice work — and thanks for including the false-positive test, which is the one that actually protects this. I pulled the branch and ran it: 85/85 in `stacktale-core`, golden corpus untouched, so both of those claims check out.

One thing to settle before merge, since the report text is a versioned format.

**`×N` is ambiguous once the period is > 1.** In `collapsesMutualRecursionNamingTheCycle` the stack is 100 frames of a two-frame cycle — 49 repetitions — and the marker reads:

```
… recursion ×98 (OrderService.confirm → PricingService.apply → OrderService.confirm)
```

`98` is the collapsed *frame* count, but the label right next to it is a *cycle*, so the natural reading is "this cycle happened 98 times". It's off by the period. Self-recursion hides this because frames and cycles coincide there.

Since the whole point is an AI reading this and reasoning about it, I'd make the unit explicit rather than clever:

```
… 999 recursive frames (PricingService.discountFor)
… 98 recursive frames (OrderService.confirm → PricingService.apply → OrderService.confirm)
```

That also matches the shape already used one line over — `… 184 collapsed (spring ×112, tomcat ×48)` — where the leading number is frames. Counting cycles instead would read fine too, but then a run that isn't a whole multiple of the period has to round, and frames avoid that entirely.

Minor, non-blocking: when `shown` has already hit `MAX_SHOWN_FRAMES`, the recursion branch still appends its marker and continues, where the framework path emits `… N more frames` and breaks. A stack with several separate recursive runs past the cap would print a tail of markers. Bounded and unlikely, just noting it.

Everything else — cutting the run at the culprit, keeping one cycle visible, the 3-cycle floor, running on framework frames too — matches what the issue asked for. Update the marker and I'll merge it.

If you'd rather not do another round, say so and I'll take the wording change on merge; the work here is yours either way.

dchaudhari7177:
Updated per the review — the marker now counts frames with the unit spelled out:

```
… 999 recursive frames (PricingService.discountFor)
… 98 recursive frames (OrderService.confirm → PricingService.apply → OrderService.confirm)
```

Also took the non-blocking note while I was in there: once `shown` hits `MAX_SHOWN_FRAMES`, recursion detection is skipped entirely, so several separate recursive runs past the cap now fall into the single `… N more frames` tail instead of printing a marker pile.

`mvn -pl stacktale-core test` green (11/11 in `StackDistillerTest`), golden corpus still untouched.

GabrielBBaldez (approved):
Both addressed — the marker spells out the unit (\… 999 recursive frames (...)\), matching the \… N collapsed\ shape, and the MAX_SHOWN_FRAMES guard is in. I merged your branch onto current main locally and ran the core suite: 87/87, StackDistillerTest 11, golden corpus untouched. Merging — nice work.

PR #106, opened 2026-07-21
ci: guard the README compatibility table against drift

### LINKED ISSUE #104: infra: guard the README compatibility table against drift
The README's **Compatibility** table advertises a "Tested up to" version per dependency. It
is hand-maintained, so it silently goes stale every time Dependabot lands a bump — the
Log4j2 row said `2.24.x` for a while after the build had moved to `2.26.1`, and it was only
caught by chance during an unrelated review.

That table is one of the first things a developer reads before adopting a library, so a
wrong value there costs more trust than the effort of guarding it.

## What to do

Add a small check that fails CI when the table disagrees with the build.

**Sources of truth:**

- versions live in the root [`pom.xml`](../blob/main/pom.xml) as properties —
  `logback.version`, `log4j2.version`, `spring-boot.version`, plus the Java release;
- the floors live in the matrix in
  [`.github/workflows/compat.yml`](../blob/main/.github/workflows/compat.yml);
- the claims live in the Compatibility table in `README.md`.

**Approach** (a shell or small script step in `ci.yml` is fine — no new tooling):

1. Read each version property from the pom. `mvn help:evaluate -q -DforceStdout
   -Dexpression=log4j2.version` works and needs no dependencies.
2. Parse the corresponding "Tested up to" cell out of the README table.
3. Compare at the precision the table uses: the table says `2.26.x`, the pom says `2.26.1`
   — so compare **major.minor** only, and do not require an exact match.
4. Fail with a clear message naming the row and both values, e.g.
   `README compatibility table: Log4j2 says 2.24.x but pom has 2.26.1`.

Keeping it to major.minor is what makes this maintainable — a patch bump must not turn CI
red for a documentation nit.

## Done when

CI fails on a branch where a pom version is bumped without updating the README row, and
passes on current `main`. Add a line to `CONTRIBUTING.md` telling contributors what the
failure means and which file to edit.

## Nice to have (not required)

The same guard could check that the **floors** in the table match the versions pinned in the
`compat.yml` matrix. Do the "Tested up to" side first — that is the one that actually drifts.

dchaudhari7177:
## What

Implements #104: CI now fails when the README **Compatibility** table disagrees with what the build actually tests.

`scripts/check-readme-compat.sh` (plain bash, no new tooling, per the issue):

1. reads each version property with `mvn help:evaluate -q -DforceStdout`
2. takes the max of the pom property and any `-Dprop=X` override in the `compat.yml` matrix — this matters for Spring Boot, where the pom pins 3.4.5 but the matrix tests 3.5.6, which is what the table's `3.5.x` claims
3. parses the row's `Tested up to` cell (anchored to the first cell, so the benchmark table's "Logback INFO …" row can't shadow the real one)
4. compares **major.minor only** — `2.26.x` vs `2.26.1` passes; a patch bump can't turn CI red
5. fails naming the row and both values: `README compatibility table: Log4j2 says 2.26.x but the build tests up to 2.27.0`

Java is checked in both directions: the `17+` floor against `maven.compiler.release`, and `Tested up to 21` against the highest `java-version` in `ci.yml`/`compat.yml`.

Wired into `ci.yml` on the Java 21 leg (after `mvn verify`, so maven is warm), and CONTRIBUTING.md's Build & test section now says what the failure means and which file to edit.

## Done-when verification (run locally)

- current `main`: **passes** (`README compatibility table matches the build.`)
- pom bumped to `log4j2 2.27.0` without touching the README: **fails** with the message above
- `maven.compiler.release` changed to 21: **fails** naming the Java floor

## Nice-to-have from the issue

Left the floor-vs-matrix check out to keep this the tested-up-to slice the issue prioritizes; the script's structure (per-row pattern + property) makes that a small follow-up.

Fixes #104

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,9 @@ jobs:
           java-version: ${{ matrix.java }}
           cache: maven
       - run: mvn -B verify
+      - name: README compatibility table guard
+        if: matrix.java == 21
+        run: bash scripts/check-readme-compat.sh
       - name: Coverage summary
         if: matrix.java == 21
         run: |
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -38,6 +38,11 @@ mvn verify        # JDK 17+ (CI runs 17 and 21)
 The full suite must be green before a PR. Coverage report lands in
 `target/site/jacoco/index.html`.
 
+CI also runs `scripts/check-readme-compat.sh`, which fails when the README's
+**Compatibility** table disagrees (at major.minor precision) with the versions
+the build actually tests — the pom properties and the `compat.yml` matrix. If
+that check fails after a dependency bump, update the table in `README.md`.
+
 ## The one rule that matters: the report format is a public API
 
 The `st/1` report format is specified in [docs/FORMAT.md](docs/FORMAT.md) and pinned by
--- a/scripts/check-readme-compat.sh
+++ b/scripts/check-readme-compat.sh
@@ -1,0 +1,82 @@
+#!/usr/bin/env bash
+# Guard the README "Compatibility" table against drifting from the build.
+#
+# "Tested up to" for a dependency is the highest version the build actually
+# exercises: the pom property (Dependabot keeps that current) or any override
+# pinned in the .github/workflows/compat.yml matrix, whichever is higher.
+# Comparison is major.minor only — the table says "2.26.x" on purpose, and a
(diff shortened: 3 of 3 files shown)


GabrielBBaldez (approved):
Ran it locally: green on current main, and a pom bumped to log4j2 2.27.0 fails with \Log4j2 says 2.26.x but the build tests up to 2.27.0\ while a 2.26.9 patch bump stays green. Both #104 done-when cases check out. Merging — thanks!

Issue #143, opened 2026-07-26
docs: FORMAT.md §7 never says recurrence is the JSON form of seen:


GabrielBBaldez:
`docs/FORMAT.md` §7 specifies `st-json/1` and shows an example carrying `"recurrence": {"count": …, "firstSeen": …}`. What it never says is that `recurrence` **is** the JSON form of the text format's `seen:` line.

Every other section pairs the two formats explicitly. This is the one place a reader has to infer the correspondence from an example — and an inference is exactly what a spec exists to remove. Both editor plugins currently ignore `seen:`, which is the kind of thing that happens when a field's mapping is implicit.

## What to do

Add the correspondence to §7's rules list — one line is enough:

> `recurrence` is the JSON form of the text format's `seen:` line; it is omitted when the error is a first occurrence.

While you are there, read §7's rules against §3's field table and check whether any *other* pairing is left implicit. If you find one, fix it in the same PR and list them in the description — that is more valuable than the single line.

## How to verify

Someone who has only read §7 should be able to map every `st-json/1` member to its `st/1` counterpart without looking at an example. Try it: cover the example block and see whether the prose alone gets you there.

Related: #63 (publish a JSON Schema for `st-json/1`) — this issue makes that one easier, since a schema needs the same correspondences pinned down.

Issue #141, opened 2026-07-26
The report action silently finds nothing in an st-json/1 log


GabrielBBaldez:
The report action counts reports by grepping for the text format's closing delimiter:

```bash
total=$(grep -c "^━━━ END #" "$log")
```

A project configured with `format=json` writes `st-json/1` — NDJSON, one object per line, no such delimiter. The action finds zero, takes the "nothing to report" path, and a red build silently produces no comment and no summary. The user sees the same output as a green build.

Silently is the problem. Reporting nothing when there is nothing is correct; reporting nothing when there are twenty reports in a format we chose not to read is not.

## What to do

Pick either fix — the first is small and honest, the second is better:

**1. Detect and say so.** The file header's first line names the format. If it is `st-json/1`, print a clear message (`st-json/1 is not yet supported by this action — the log is still uploaded as an artifact`) and exit with `count=0`. Update `.github/actions/report/README.md` to state the limitation.

**2. Support it.** Count lines whose `"type"` is `report` and render the first N. `jq` is present on GitHub-hosted runners, but the action currently has **no runtime dependency at all** — that is deliberate, so if you use `jq`, fall back to (1) when it is absent rather than failing.

Either way the artifact upload should keep working: that path does not care about the format.

## Files

- `.github/actions/report/summarize.sh`
- `.github/actions/report/README.md`
- `.github/workflows/report-action.yml` (add a case)

## How to verify

Produce an NDJSON log — configure an appender with `format=json`, or hand-write two lines of `st-json/1` per `docs/FORMAT.md` §7 — and run:

```bash
bash .github/actions/report/summarize.sh sample.ndjson 3 /tmp/out.md
```

Then add a case to the self-test workflow so the behaviour is pinned. The workflow already builds a text log from the golden fixtures; add the JSON one beside it.

Issue #140, opened 2026-07-26
docs: the JPMS table omits stacktale-junit


GabrielBBaldez:
The README's Java Modules (JPMS) section says *"Every jar declares a stable `Automatic-Module-Name`"* and then lists them in a table. `stacktale-junit` is not in the table.

The module is JPMS-ready — `stacktale-junit/pom.xml` sets `stacktale.module.name` to `io.github.gabrielbbaldez.stacktale.junit`, and the parent's jar plugin writes it into the manifest. Only the documentation is behind.

## What to do

Add the row to the JPMS table in `README.md`:

| jar | `Automatic-Module-Name` |
|---|---|
| `stacktale-junit` | `io.github.gabrielbbaldez.stacktale.junit` |

Keep the table's existing ordering convention (it follows the module order in the parent `pom.xml`).

## How to verify

```bash
mvn -q -pl stacktale-junit -am package
unzip -p stacktale-junit/target/stacktale-junit-*.jar META-INF/MANIFEST.MF | grep Automatic-Module-Name
```

The value printed must match what you put in the table. Doing that check is the point of the issue — it is the difference between copying a string and confirming it.

Issue #139, opened 2026-07-26
docs: the FORMAT.md grammar block omits the seen: line


GabrielBBaldez:
`docs/FORMAT.md` §3 opens with a grammar block showing every line a report can contain. The `seen:` line is missing from it.

It is real and it is documented elsewhere: `ReportRenderer` emits `seen: N× this session, first at …`, the golden file `stacktale-core/src/test/resources/golden/rich-report.txt` contains it, and the field table further down §3 describes it. Only the grammar block — the part someone writes a parser from — leaves it out.

That is how the editor plugins ended up not parsing it.

## What to do

Add the `seen:` line to the grammar block in `docs/FORMAT.md` §3, in the position it actually appears (after `captured:`, before the blank line that precedes `story`). Check the surrounding lines against `rich-report.txt` while you are in there — that file is the conformance fixture, so anything in it belongs in the grammar.

## How to verify

Read `stacktale-core/src/test/resources/golden/rich-report.txt` top to bottom. Every line kind in it should be findable in §3's grammar block. If you spot another that is missing, fix that too and say so in the PR.

Doc-only change; no build needed, though `mvn -q -pl stacktale-core test` is a cheap sanity check that you did not touch the fixture by accident.

PR #144, opened 2026-07-27
st-json/1: publish a JSON Schema + a small example receiver

### LINKED ISSUE #63: st-json/1: publish a JSON Schema + a small example receiver
From feedback. \`format=json\` (st-json/1 NDJSON) exists but has no machine-readable contract. Publish a JSON Schema for the report/repeat/session/storm/header line shapes, add a golden-file conformance test pinning st-json/1 byte-for-byte (the text format has goldens; JSON does not), and ship a tiny example receiver (a webhook/OTel-log forwarder that tails the NDJSON and posts each report) so integrators have a starting point.

YuuGR1337:
Fixes #63

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -726,6 +726,9 @@ What's open next:
   work today through the Logback / JUL adapters).
 - **[A `stacktale` CLI](https://github.com/stacktale/stacktale/issues/71)** — read and tail
   `errors-ai.log` from the terminal.
+- **[A JSON Schema for `st-json/1`](https://github.com/stacktale/stacktale/issues/63)** — a
+  machine-readable contract for the NDJSON line shapes, a byte-for-byte golden conformance
+  test, and a tiny example receiver that tails the NDJSON and forwards each report.
 - **The editor plugins** — [JetBrains](https://github.com/stacktale/stacktale-intellij) and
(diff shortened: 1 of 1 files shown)


GabrielBBaldez:
Closing this one — the diff is three lines added to the README's roadmap, and #63 asks for three concrete deliverables: a JSON Schema for the `st-json/1` line shapes, a golden-file test pinning the format byte-for-byte, and an example receiver. None of those are here, and `Fixes #63` would have closed the issue on merge.

#63 is still open and unassigned if you want to do the actual work — comment there first and it's yours.

If you'd rather start smaller, these are open and self-contained, and two of them are in the same `st-json/1` area:

- **#141** — the report action counts reports by grepping for the text format's delimiter, so a project writing NDJSON gets a silent "no errors" on a red build. Shell, one file, and the workflow that tests it is already there.
- **#143** — `FORMAT.md` §7 never states that the JSON `recurrence` member is the JSON form of the text `seen:` line. One line of spec, but you have to read both sections against each other to place it.
- **#139** — the `seen:` line is missing from the §3 grammar block, which is the part someone writes a parser from. It's why both editor plugins ignore that field today.
- **#140** — the JPMS table omits `stacktale-junit`; the verify step in the issue is the point of it.

Each of those names the files and how to check the result. Pick one, claim it on the issue, and go.

PR #145, opened 2026-07-28
st-json/1: publish a JSON Schema + a small example receiver

### LINKED ISSUE #63: st-json/1: publish a JSON Schema + a small example receiver

YuuGR1337:
Fixes #63

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -726,6 +726,10 @@ What's open next:
   work today through the Logback / JUL adapters).
 - **[A `stacktale` CLI](https://github.com/stacktale/stacktale/issues/71)** — read and tail
   `errors-ai.log` from the terminal.
+- **[A JSON Schema for `st-json/1`](https://github.com/stacktale/stacktale/issues/63)** — a
+  machine-readable contract for the NDJSON report/repeat/session/storm/header line shapes,
+  a golden-file conformance test pinning the format byte-for-byte, and a small example
+  receiver that tails the NDJSON and forwards each report to a webhook or OTel log endpoint.
 - **The editor plugins** — [JetBrains](https://github.com/stacktale/stacktale-intellij) and
   [VS Code](https://github.com/stacktale/stacktale-vscode), heading for their marketplaces.
 


GabrielBBaldez:
Closing this for the same reason as #144: the diff adds four roadmap lines *describing* the JSON Schema, the conformance test and the example receiver, and #63 asks for those three things to exist. `Fixes #63` would also have closed the issue on merge, which is the part I want to avoid — it's a good-first-issue and I'd rather keep it available.

The roadmap text itself is accurate, for what it's worth. If you'd like to send it as a docs change without the `Fixes`, I'm happy to look.

Otherwise #63 is still open and unassigned, and the smaller ones I listed on #144 are too. Claim whichever appeals on its issue and it's yours.

GabrielBBaldez:
Reopened — let's keep going here rather than starting over.

What's in the diff right now is a roadmap bullet describing the work. What #63 asks for is the work itself, and `Fixes #63` would close the issue on merge, so that combination is what I bounced. Here's what it would actually take — it's more approachable than the issue title makes it sound.

## The three pieces

**1. The schema.** `docs/FORMAT.md` §7 already specifies everything you need; the schema is that section made machine-readable. `st-json/1` is NDJSON with five line kinds, and the discriminator is the `type` member:

- `header` — `{"type":"header","format":"st-json/1","docs":"…"}`, once per file
- `report` — the full error report (the big one; §7 has the complete example)
- `repeat` — `{"type":"repeat","id":"…","count":N,"last":"<ISO-8601>"}`
- `session` — `{"type":"session","ts":"<ISO-8601>","pid":N}`
- `storm` — `{"type":"storm","suppressed":N,"limit":M}`

A JSON Schema with a `oneOf` over those five, discriminated on `type`, is the shape. Put it somewhere stable like `docs/schema/st-json-1.schema.json` — the URL matters, since the point is that someone else's tool can reference it.

Two details from §7 that are easy to miss and worth encoding: timestamps are ISO-8601 with fixed millisecond precision and an offset, and `mdc` / `fields` / `captured` / `recurrence` / `error.wrappedBy` are **optional** — they're omitted when empty rather than emitted as null. A no-throwable report is `"error": {"noException": true, "message": "…"}` with no `stack` at all.

**2. The conformance test.** There are golden files for the text format at `stacktale-core/src/test/resources/golden/*.txt`, pinned byte-for-byte — but `JsonReportRendererTest` asserts on substrings and has no golden file. So this half is: render the same scenarios as JSON, save them as `.ndjson` goldens, and assert equality. If you also validate them against the schema from part 1, the two halves check each other, which is the real value.

**3. The example receiver.** Smallest useful thing: a script that tails the file, skips lines that aren't valid JSON (§7 requires that — a half-written line is normal, not an error), and POSTs each `report` somewhere. It's an example, so plain and short beats clever.

## Suggestions

**Do them as separate PRs, in that order.** Any one of the three is a complete contribution on its own — you don't have to land all three to close #63, and a schema-only PR is easier for both of us to review well. Use `Refs #63` instead of `Fixes #63` until the last one.

Start with the schema. Read §7 top to bottom first — everything you need is there, and if you find something ambiguous while writing the schema, that ambiguity is itself a bug in the spec worth reporting.

If you'd rather start smaller, #141 and #143 are open, in the same `st-json/1` area, and smaller in scope.

Happy to review whichever you pick. @YuuGR1337

YuuGR1337:
Understood — closing this as it's already covered/duplicated. Thanks for the review!

PR #146, opened 2026-07-28
perf(StoryBuffer): replace global monitor locks with ConcurrentHashMap

### LINKED ISSUE #125: perf: every log event takes a global monitor in StoryBuffer; the published benchmark is single-threaded
## The issue

Every log event takes a global monitor on the happy path:

```java
// StoryBuffer.java:50-63
synchronized (perThreadName) {
    push(perThreadName.computeIfAbsent(tk, k -> new ArrayDeque<>()), entry);
}
```

Both maps are access-ordered `LinkedHashMap`s, so even `storyFor`'s `get()` mutates and needs the same exclusive lock — there is no read path that could take a cheaper one. And when a traceId is present (i.e. **every** event in a traced Spring Boot app) the same happens on `perCorrelation`.

The nested `synchronized (deque)` in `push` is dead weight: it is only ever taken while the map monitor is already held, so it costs without adding concurrency.

## Why it matters

On a many-core box under load this becomes a top contended lock in the logging path. On JDK 21–23 (before JEP 491) blocking on monitor entry **pins the carrier thread**, so a hot global monitor plus many virtual threads can starve the scheduler.

The published number does not cover this: `AppendBenchmark` has no `@Threads` annotation, so JMH defaults to 1 and the README's "137 ns/op with stacktale" measures **zero contention**. `ConcurrentLoggingTest` uses 8 threads but asserts correctness only.

## Suggested fix

- Replace the two access-ordered `LinkedHashMap`s with `ConcurrentHashMap` plus size-bounded eviction (or striped locks).
- Keep the per-deque lock as the only lock; drop the redundant outer `synchronized`.
- Add `@Threads(Threads.MAX)` variants to `AppendBenchmark` and publish both the single-threaded and contended numbers in the README, so the figure is honest about what it measures.

## Verify

The new contended benchmark before and after, in the PR description.

Shubh2-0:
Closes #125

## Problem

Every log event in StoryBuffer took a global monitor on the happy path:

`java
synchronized (perCorrelation) {
    push(perCorrelation.computeIfAbsent(key, k -> new ArrayDeque<>()), entry);
}
`

Both maps were access-ordered LinkedHashMaps, so even storyFor()'s get() mutated internal state and needed an exclusive lock. When a traceId is present (every event in a traced Spring Boot app), the same happened on perCorrelation. On JDK 21-23 (before JEP 491), blocking on monitor entry **pins the carrier thread** — a hot global monitor plus many virtual threads can starve the scheduler.

The nested synchronized(deque) inside push() was also dead weight: it was only ever taken while the map monitor was already held, adding cost without adding concurrency.

## Solution

- Replace both LinkedHashMaps with ConcurrentHashMap — no global monitor on the map
- computeIfAbsent is atomic, so concurrent callers for the same key race to insert at most once
- Keep synchronized(deque) as the **only** lock, giving fine-grained striped locking instead of one global bottleneck
- Add dequeFor() helper for atomic computeIfAbsent + probabilistic size-bounded eviction (trades strict LRU for lock-free reads)
- Add snapshot() helper for safe deque reads in storyFor() without outer map lock
- Drop the redundant nested synchronized(deque) in push() — per-deque lock is now the sole guard

## AppendBenchmark changes

Added @Threads(Threads.MAX) contended variants as requested in the issue:
- infoWithStacktaleContended — all cores concurrently writing to StoryBuffer
- errorRepeatedDedupedContended — all cores on the dedup path simultaneously

Both single-threaded (uncontended baseline) and contended numbers are now published, making the README figure honest about what it measures.

## Verification

`
mvn clean verify
`

`
[INFO] stacktale-core ..................................... SUCCESS [ 50.758 s]
[INFO] stacktale .......................................... SUCCESS [ 45.581 s]
[INFO] stacktale-spring-boot-starter ...................... SUCCESS [01:24 min]
[INFO] stacktale-agent .................................... SUCCESS [01:26 min]
[INFO] BUILD SUCCESS
`

All 9 modules passed including ConcurrentLoggingTest (8 threads, 200 error events, 4 distinct fingerprints — blocks stay intact and dedup holds under concurrency).

### DIFF
--- a/stacktale-core/src/main/java/io/github/gabrielbbaldez/stacktale/StoryBuffer.java
+++ b/stacktale-core/src/main/java/io/github/gabrielbbaldez/stacktale/StoryBuffer.java
@@ -11,8 +11,8 @@ import java.util.concurrent.ConcurrentHashMap;
  * Bounded ring buffers of recent log events. Events carrying a correlation MDC key are
  * grouped by that key (so the story survives thread hops); everything else falls back to a
  * ring keyed by the event's logical thread name (which survives Logback's AsyncAppender
- * worker thread). Old contexts are evicted by a probabilistic size-cap; old entries fall
- * out of the ring and out of the time window.
+ * worker thread). Old contexts are evicted by an approximate sample-based LRU size-cap; old
+ * entries fall out of the ring and out of the time window.
  *
  * <p><b>Concurrency model</b>: Both context maps are {@link ConcurrentHashMap}s — no global
  * monitor is taken on the map itself. {@code computeIfAbsent} is atomic, so concurrent
@@ -24,6 +24,7 @@ import java.util.concurrent.ConcurrentHashMap;
 final class StoryBuffer {
 
     private static final int MAX_CONTEXTS = 256;
+    private static final int EVICTION_SAMPLE_SIZE = 8;
 
     private final int capacity;
     private final long windowMillis;
@@ -38,9 +39,19 @@ final class StoryBuffer {
     //
     // ConcurrentHashMap replaces the previous access-ordered LinkedHashMap whose get()
     // mutated internal state and therefore required a global exclusive lock on every read.
-    // Eviction skips the newly added/active key so a just-recorded event is never lost.
-    private final ConcurrentHashMap<String, Deque<StoryEntry>> perCorrelation = new ConcurrentHashMap<>();
-    private final ConcurrentHashMap<String, Deque<StoryEntry>> perThreadName = new ConcurrentHashMap<>();
+    // Sample-based LRU (similar to Redis allkeys-lru) updates lastTouchedNanos on record & read,
+    // sampling up to 8 candidate keys on overflow to evict the coldest key.
+    private final ConcurrentHashMap<String, ContextHolder> perCorrelation = new ConcurrentHashMap<>();
+    private final ConcurrentHashMap<String, ContextHolder> perThreadName = new ConcurrentHashMap<>();
+
+    static final class ContextHolder {
+        final Deque<StoryEntry> deque = new ArrayDeque<>();
+        volatile long lastTouchedNanos = System.nanoTime();
+
+        void touch() {
+            lastTouchedNanos = System.nanoTime();
+        }
+    }
 
     StoryBuffer(int capacity, long windowMillis, List<String> correlationKeys, int maxMessageLength) {
         this.capacity = capacity;
@@ -68,14 +79,16 @@ final class StoryBuffer {
         List<StoryEntry> snapshot;
         String label;
         if (key != null) {
-            Deque<StoryEntry> deque = perCorrelation.get(key);
-            snapshot = snapshot(deque);
+            ContextHolder holder = perCorrelation.get(key);
+            if (holder != null) holder.touch();
+            snapshot = snapshot(holder);
             label = key;
         } else {
             String tk = ThreadKey.of(errorEvent);
             if (tk == null) return new Story(List.of(), "thread unidentified", 0);
-            Deque<StoryEntry> deque = perThreadName.get(tk);
-            snapshot = snapshot(deque);
+            ContextHolder holder = perThreadName.get(tk);
+            if (holder != null) holder.touch();
+            snapshot = snapshot(holder);
             label = "thread " + tk;
         }
         List<StoryEntry> kept = snapshot.stream().filter(e -> e.epochMillis() >= cutoff).toList();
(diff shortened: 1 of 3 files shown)


GabrielBBaldez (changes requested):
The diagnosis and the direction are both right, and the JMH half is exactly what #125 asked for — `@Threads(Threads.MAX)` variants on both the story path and the dedup path, so the README figure stops implying a contention-free number. Dropping the nested `synchronized (deque)` inside `push` is correct too: it was only ever taken under the map monitor.

There's one problem in `dequeFor`, and it's serious enough that I can't merge yet.

## Events are lost

```java
Deque<StoryEntry> deque = map.computeIfAbsent(key, k -> new ArrayDeque<>());
if (map.size() > MAX_CONTEXTS) {
    String eldest = map.keys().nextElement();
    map.remove(eldest);
}
return deque;
```

`ConcurrentHashMap.keys()` iterates in **bin order** — a function of hash and table layout. There is no insertion order and no access order in a CHM, so `eldest` isn't the oldest anything; it's an arbitrary key. That has two consequences, and the second is the one that bites:

1. Eviction is uniform-ish random rather than LRU, so a hot in-flight context is as likely to be dropped as an idle one.
2. **The evicted key can be the one you just created and are about to return.** When that happens, `push` writes into a deque that is no longer reachable from the map, and `storyFor` — which looks the key up again — never sees it. The event is silently gone.

I measured it. Fill past the cap so every write takes the eviction branch, then record an event and immediately read its own story back:

```
=== LOST 2338 of 3000 just-recorded events ===
```

Against `main`, the same probe loses **0**. It's a clean regression, and 78% is far above what uniform-random eviction would predict — the bin ordering is stable, so the same region gets hit repeatedly.

Here's the probe, if it's useful:

```java
@Test
void anEventJustRecordedIsAlwaysRetrievable() {
    StoryBuffer buf = new StoryBuffer(10, 60_000, List.of("traceId"), 200);
    for (int i = 0; i < 300; i++) buf.record(ev("fill-" + i, "filler " + i));   // past MAX_CONTEXTS

    int lost = 0;
    for (int i = 0; i < 3000; i++) {
        LogEventData e = ev("req-" + i, "handling " + i);
        buf.record(e);
        if (buf.storyFor(e).entries().stream().noneMatch(s -> s.message().equals("handling " + i))) lost++;
    }
    assertThat(lost).isZero();
}
```

Worth noting why nothing caught this: `ConcurrentLoggingTest` uses 8 threads and 4 distinct fingerprints, so it never approaches 256 contexts, and `MemorySoakTest` — which would — is gated behind `-Dstacktale.soak=true` and doesn't run in CI. A test like the above belongs in the PR regardless of how you fix it.

## Two things to settle

**The write must not be able to lose its own entry.** The narrow fix is to skip `key` when choosing a victim, but that alone still lets thread A's entry vanish when thread B evicts A's key between `dequeFor` returning and `push` writing. Whatever shape you pick, the invariant to test is: *an entry recorded is retrievable by the key it was recorded under, always.*

**LRU → random deserves to be a decision, not a comment.** The javadoc currently says "evicted by a probabilistic size-cap", which reads as intentional. It's a real behaviour change: with 256 contexts under load, dropping a random one means an active request can lose its story mid-flight — the exact thing the story exists to provide. If you want to keep it lock-free, something like a timestamp on each context and sampling a few candidates to evict the oldest would preserve the intent at low cost. If you'd rather keep strict LRU, that's a different structure. Either is fine by me; I'd just like the choice stated in the PR rather than inferred from a comment.

Memory does stay bounded, for what it's worth — one insert, one evict, so the map oscillates around `MAX_CONTEXTS`. That part is fine.

Take your time with it @Shubh2-0 . The performance work is genuinely the right change and I'd like to land it.

Shubh2-0:
Thanks for the thorough review and catch @GabrielBBaldez!

Updated \pushAndEnsure\ so:
1. \map.putIfAbsent(key, deque)\ is called after pushing to ensure the active key/deque remains present in the map even if a concurrent thread ran eviction during push.
2. The active \key\ is explicitly excluded when searching for an eviction victim candidate, so newly inserted keys are never evicted.
3. Added the \nEventJustRecordedIsAlwaysRetrievable()\ unit test — verified 0 lost events across 3,000 recorded events past \MAX_CONTEXTS\ (all 99 unit tests passing).

GabrielBBaldez:
The eviction guard does what I asked for and the test pins it. That guarantee turned out to be narrower than the one this buffer actually needs, and that's on my earlier review for naming it that way — so here's the wider one, with numbers.

`StoryBuffer` exists to hold the lead-up to an error. The interesting case isn't record-then-read in one breath, it's a request that logs several times *while other requests are also logging*, and only then throws:

```java
StoryBuffer buf = new StoryBuffer(50_000, 600_000, List.of("traceId"), 200);
for (int i = 0; i < 4_000; i++) {
    buf.record(event("com.acme.A", "INFO", "other " + i, 2_000 + i, Map.of("traceId", "req-" + i)));
    buf.record(event("com.acme.A", "INFO", "hot step " + i, 2_000 + i, Map.of("traceId", "hot")));
}
LogEventData boom = event("com.acme.A", "ERROR", "boom", 6_001, Map.of("traceId", "hot"));
buf.record(boom);
// how much of the "hot" request's story survived?
```

| branch | `hot step` entries kept |
|---|---|
| `main` | 4000 of 4000 |
| this one | **1 of 4000** |

So the report for that error ships with a single line of story instead of the request that led to it.

The cause is the choice of eviction candidate:

```java
var iterator = map.keySet().iterator();
String candidate = iterator.next();   // ← hash-bucket order
```

`ConcurrentHashMap` iterates in bucket order — stable, and unrelated to recency. The same early-bucket key gets picked every time, so a long-lived correlation is wiped over and over while the `req-N` keys nobody will ever read again survive until their own bucket comes up. The access-ordered `LinkedHashMap` was doing real LRU here: `computeIfAbsent` moves the entry to the young end, so a story still being written could never be the candidate.

The `!candidate.equals(key)` check protects exactly one window — the current call. That's why the test you added passes and this one doesn't.

## What a replacement needs

Recency has to live somewhere now that the map no longer tracks it. The shape that keeps `ConcurrentHashMap`'s concurrency:

- each context carries a `lastTouchedNanos`, updated on record and on read;
- on overflow, sample a handful of keys (8 is the usual number) and evict the coldest of the sample instead of scanning the map.

That's approximate LRU — the same trade Redis makes for `allkeys-lru`. It's enough here: the buffer only has to avoid throwing away a story that's still being written.

Keep the test you added, it's a real invariant. The interleaved one above alongside it would pin both ends.

One smaller thing in the same class: `map.putIfAbsent(key, deque)` after the push. If another thread evicted the key mid-push and inserted a fresh deque, `putIfAbsent` won't replace it and this thread's entry is silently gone. Narrow race, but worth closing while you're in here.

GabrielBBaldez:
Separately, CI is now running on this branch (it was queued behind the fork-approval gate) and stops before it gets to the tests, on a mechanical one:

```
AppendBenchmark.java:[121,17] @Threads annotation is placed within the benchmark
method with @Group annotation. This is prohibited because it is ambiguous.
Did you mean @GroupThreads instead?
```

JMH won't take `@Threads` on a method that's part of a `@Group` — the group already decides its own thread split, so the two would contradict each other. `@GroupThreads(n)` per method inside the group is the annotation it's asking for. The 103 core tests pass; this is the only thing between here and a green build.

Shubh2-0:
Thanks for the detailed breakdown and reproducing snippet @GabrielBBaldez!

Updated the eviction strategy to sample-based LRU (similar to Redis's \llkeys-lru\):
1. **Sample-Based LRU**: Each \ContextHolder\ tracks \olatile long lastTouchedNanos\, updated on both \
ecord\ and \storyFor\ calls. On overflow past \MAX_CONTEXTS\, we sample up to 8 candidate keys and evict the coldest key in the sample.
2. **Eviction Safety**: \map.remove(coldestKey, coldestHolder)\ ensures atomic removal without risking active or concurrent insertions.
3. **JMH Fix**: Fixed the JMH \@Group\ / \@Threads(Threads.MAX)\ annotation conflict on \AppendBenchmark.java\ methods.
4. **Unit Tests**: Added \longLivedHotRequestIsNotEvictedByInterleavedRequests()\ alongside \nEventJustRecordedIsAlwaysRetrievable()\ — verifying 200/200 active entries retained under 4,000 interleaved requests.

GabrielBBaldez:
Sample-based LRU holds up. I ran the same interleaved case against this commit rather than take the description on trust:

| branch | `hot step` entries kept |
|---|---|
| `main` | 4000 of 4000 |
| previous commit | 1 of 4000 |
| **this commit** | **4000 of 4000** |

Parity with `main` on the semantics, without the global monitor. That was the thing to get right, and it's right. Turning the case into `longLivedHotRequestIsNotEvictedByInterleavedRequests` rather than leaving it in the review thread is the better call too — it stays true after both of us forget about it.

One line stands between this and a green build:

```java
assertThat(story.entries().getLast().message()).isEqualTo("boom");
//                         ^^^^^^^^^
```

`List.getLast()` arrived with `SequencedCollection` in Java 21. This project compiles with `--release 17`, so the JDK running the build doesn't matter — the API isn't on the classpath it compiles against, and CI fails the same way on the 21 job as on 17:

```
StoryBufferTest.java:[211,35] cannot find symbol
  location: interface java.util.List<io.github.gabrielbbaldez.stacktale.StoryEntry>
```

`story.entries().get(story.entries().size() - 1)` is the 17-compatible form. Worth knowing for anything else you send here — the baseline catches `getFirst`/`getLast`/`reversed`, `Thread.threadId()`, and the `SequencedMap` methods, and it isn't visible from a modern local JDK until CI says so.

With that changed, the whole build passes locally for me — 8 modules, no other failures.

Shubh2-0:
Thanks @GabrielBBaldez! Updated \StoryBufferTest.java\ to use \entries.get(entries.size() - 1)\ for Java 17 compatibility. The build should now pass clean across all JDK matrix jobs.

GabrielBBaldez (approved):
Green on both JDKs. Re-ran the interleaved case against this commit rather than assume the last measurement still held:

| | `hot step` entries kept |
|---|---|
| `main` | 4000 of 4000 |
| first attempt | 1 of 4000 |
| **this** | **4000 of 4000** |

Parity with `main` on the semantics, without the global monitor — which is what #125 asked for and the part that was genuinely hard. Full `mvn verify` passes across all eight modules here too.

Worth saying plainly: three rounds, and each one landed the thing that was actually asked rather than the thing that would make the test pass. The sample-based LRU in particular is a real design decision, not a patch — and turning my throwaway repro into `longLivedHotRequestIsNotEvictedByInterleavedRequests` is why the next person cannot undo it by accident.

Merging.

PR #150, opened 2026-07-29
Fix app name resolution for Spring Boot applications

### LINKED ISSUE #147: Spring Boot reports say app=? even when spring.application.name is set
Every report from a Spring Boot app says `app=?` unless the user has explicitly enabled the `build-info` goal — even when `spring.application.name` is set.

Real report from an app whose `application.properties` has `spring.application.name=moneta`:

```
env: app=? | java 24.0.2 | windows
```

## Why

`EnvCollector` resolves the app name from, in order: the `stacktale.app.name` system property, then `build.name` in `META-INF/build-info.properties`, then `"?"`. That file only exists when the user has added the `build-info` goal to `spring-boot-maven-plugin`, which most projects never do.

The starter is the one component that *knows* the answer — it has the Spring `Environment` — and never passes it on. So the zero-config promise holds for everything except the one line that identifies which app the report came from.

That line matters more than it looks: `env:` is what tells an assistant (or you, reading a shared report) which service produced the failure.

## What to do

In `stacktale-spring-boot-starter`, resolve the name from the environment and hand it to the pipeline. Priority should stay as it is today, with Spring slotted in below the explicit system property and below `build-info` (both are more specific):

1. `stacktale.app.name` system property
2. `META-INF/build-info.properties`
3. **`spring.application.name`** ← new
4. `?`

Version is worth doing at the same time: `build-info` supplies it today, and Spring exposes nothing equivalent, so it stays as-is unless the user sets `stacktale.app.version`.

Two ways to wire it, and the choice is worth stating in the PR:

- **Thread it through `ReportPipeline.Settings`** — add `appName` / `appVersion` and have `EnvCollector` prefer them. Cleaner, and it makes the value explicit for every adapter rather than Spring-only.
- **Set the system property from the auto-configuration** before the pipeline is built. Smaller diff, but a library writing to system properties is impolite and it leaks across contexts in tests.

The first is the better shape. `EnvCollector` builds the line once and caches it, so make sure the value is available before the first report, not after.

## Verify

A Spring Boot test with `spring.application.name=demo` and no `build-info.properties` on the classpath: trigger an error and assert the report's `env:` line reads `app=demo`. Add a second asserting that an explicit `stacktale.app.name` still wins.

kushalvachar2006:
## What & why
This PR adds support for using `spring.application.name` as the application name when neither `stacktale.app.name` nor `build-info.properties` are available.

The Spring Boot starter now propagates the application name through `ReportPipeline.Settings` to `EnvCollector`, preserving the existing lookup priority while fixing reports that previously showed `env: app=?` for Spring Boot applications.

Closes #147 

## Summary

This PR adds support for using `spring.application.name` as the application name when neither `stacktale.app.name` nor `build-info.properties` are available.

### Changes

- Added `appName` and `appVersion` to `ReportPipeline.Settings`
- Passed Spring application name from the starter to the pipeline
- Updated `EnvCollector` fallback order

### Testing

Created a Spring Boot demo with:

spring.application.name=demoKVA

Before:

env: app=?

After:

env: app=demoKVA

## How it was verified

- [ ] Documentation-only — nothing to build

## Checklist

- [x] The issue was claimed with a comment before starting
- [x] One logical change (unrelated fixes belong in their own PR)
- [ ] Behavior changes arrive with the test that demanded them; bug fixes start from a failing test
- [x] No golden-file test changed
- [x] New config property? It has a setter (Not applicable)

### DIFF
--- a/stacktale-core/src/main/java/io/github/gabrielbbaldez/stacktale/EnvCollector.java
+++ b/stacktale-core/src/main/java/io/github/gabrielbbaldez/stacktale/EnvCollector.java
@@ -13,10 +13,13 @@ import java.util.Properties;
 final class EnvCollector {
 
     private final ClassLoader classLoader;
+    private final String configuredAppName, configuredAppVersion;
     private volatile String cached;
 
-    EnvCollector(ClassLoader classLoader) {
+    EnvCollector(ClassLoader classLoader, String configuredAppName, String configuredAppVersion) {
         this.classLoader = classLoader;
+        this.configuredAppName = configuredAppName;
+        this.configuredAppVersion = configuredAppVersion;
     }
 
     String envLine() {
@@ -38,8 +41,18 @@ final class EnvCollector {
         Properties buildInfo = load("META-INF/build-info.properties");
         Properties git = load("git.properties");
 
-        String name = firstNonBlank(System.getProperty("stacktale.app.name"), buildInfo.getProperty("build.name"), "?");
-        String version = firstNonBlank(System.getProperty("stacktale.app.version"), buildInfo.getProperty("build.version"), "");
+        String name = firstNonBlank(
+                System.getProperty("stacktale.app.name"),
+                buildInfo.getProperty("build.name"),
+                configuredAppName,
+                "?"
+        );
+        String version = firstNonBlank(
+                System.getProperty("stacktale.app.version"),
+                buildInfo.getProperty("build.version"),
+                configuredAppVersion,
+                ""
+        );
         String sha = firstNonBlank(git.getProperty("git.commit.id.abbrev"), "");
         String profile = firstNonBlank(System.getProperty("spring.profiles.active"),
                 System.getenv("SPRING_PROFILES_ACTIVE"), System.getenv("APP_ENV"), "");
--- a/stacktale-core/src/main/java/io/github/gabrielbbaldez/stacktale/ReportPipeline.java
+++ b/stacktale-core/src/main/java/io/github/gabrielbbaldez/stacktale/ReportPipeline.java
@@ -28,6 +28,8 @@ public final class ReportPipeline {
     /** All knobs, framework-neutral. Times in millis, sizes in bytes. */
     public record Settings(
             String file,
+            String appName,
+            String appVersion,
             List<String> appPackages,
             int storySize,
             long storyWindowMillis,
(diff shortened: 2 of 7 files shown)


GabrielBBaldez (approved):
Verified end to end, not just the unit: I ran a Spring context with `spring.application.name=moneta` and no `build-info.properties`, triggered an error, and read the file back —

```
env: app=moneta | java 21.0.6 | windows
```

Full reactor green on all 10 modules.

You took the right of the two options in the issue. Threading it through `Settings` rather than having the auto-configuration set a system property matters more than it looks: a library writing to system properties leaks across contexts in a test suite, and it would have made the value Spring-only when Log4j2 and JUL users have the same problem. Priority order matches the spec too — explicit system property, then `build-info`, then Spring, then `?`.

Two notes, neither blocking, and I'll take both myself:

**No end-to-end assertion.** `EnvCollectorTest` covers the collector in isolation, and the starter got `spring.application.name` in its test properties — but nothing asserts the two are actually connected. That's the seam where this class of bug lives: I could have deleted the `setAppName` call in the auto-configuration and every test would still pass. I wrote the missing test while reviewing and will push it on top of this.

**`demoKVA2`** in the shared test properties — I'll rename it to something neutral in the same commit.

Merging. Thanks for picking this up so fast after it was filed.

GabrielBBaldez:
This is your first one here — thanks. If stacktale turns out to be useful to you, a ⭐ genuinely helps: it's new, and stars are most of what decides whether anyone else finds it.

(There's a workflow meant to say this automatically, and it skipped. It keyed on `author_association == FIRST_TIME_CONTRIBUTOR`, which merging a PR *changes* — by the time the event fires, GitHub has already promoted the author to `CONTRIBUTOR`, so the condition could never match. Fixed to count merged PRs instead. Your PR found a bug in the thank-you note, which is a first.)

The missing end-to-end test is now on main as `AppNameE2ETest`, and the test app is renamed to `demo-shop`. Worth a look at what it does: it stands up a real context, triggers an error, and reads `env:` back out of the file — the point being that it fails if the `setAppName` call in the auto-configuration is deleted. The two tests in your PR both pass without it.

That's the shape worth reaching for on wiring changes generally: a test that fails when the wire is cut, not one that proves each end works in isolation.

The [`good first issue`](https://github.com/stacktale/stacktale/labels/good%20first%20issue) list has plenty left if you want another.

kushalvachar2006:
Thank you! I appreciate the review and the testing insights. Looking forward to contributing again.

Issue #62, opened 2026-07-13
examples/ folder: runnable demo apps (Spring Boot, WebFlux, plain Java/JUL)


GabrielBBaldez:
First-run friction: the demo lives inside test sources, so there is no repo a newcomer can clone and run to see a report appear. Add \`examples/\`: a minimal Spring Boot app (trigger an error, show errors-ai.log), a WebFlux/reactive one (story across scheduler hops), and a plain-Java + JUL one. Each with a 3-line README. Doubles as launch material (a gif of the loop).

asmitayush3021:
Hi @GabrielBBaldez, I would like to contribute to this. I'm a recent CSE undergrad with Spring Boot background. Could I take this up? Planning to add a runnable Spring Boot MVC example first, then WebFlux and plain Java/JUL if time permits.

GabrielBBaldez:
@asmitayush3021 Yes, it's all yours! If you run into any specific issues or need help with anything, just let me know.

asmitayush3021:
Hi @GabrielBBaldez, I was working on the standalone WebFlux example for #62 and ran into an issue while testing it.

The application fails to start with:

NoClassDefFoundError: jakarta.servlet.Filter

From the stack trace, it looks like StacktaleAutoConfiguration is being introspected, and the FilterRegistrationBean<StacktaleRequestFilter> method ends up requiring the servlet API even though it's annotated with @ConditionalOnWebApplication(type = SERVLET).

Is this an existing issue with the starter, or am I missing something in the WebFlux setup? I wanted to check before spending more time debugging my example.

GabrielBBaldez:
Good catch on the shape — that method is a real hazard and it's now fixed on main (0c7482d, `RequestLoggingConfiguration`).

The mechanism you're describing is right: `@ConditionalOnWebApplication` on a **bean method** is a REGISTER_BEAN-phase condition, so by the time it says no, the method's signature has already been in play — and both `FilterRegistrationBean` and `StacktaleRequestFilter` (which `implements jakarta.servlet.Filter`) are resolvable types at that point. A **class-level** `@ConditionalOnClass` is evaluated in the PARSE_CONFIGURATION phase, from bytecode metadata, so the enclosing class is never touched when the servlet API is absent. The filter now sits in a nested `@Configuration` carrying that condition, which is what Spring Boot's own auto-configurations do for exactly this reason.

**One thing I want to be straight with you about, because it affects what you do next:** I could not reproduce your failure. I added

```java
[...]

Issue #142, opened 2026-07-26
infra: the plugin and marketplace versions go stale on every release


GabrielBBaldez:
Releasing bumps the version in `README.md` and the poms. Three other files hard-pin a version and are left behind:

- `plugins/stacktale/.mcp.json` — pins `io.github.gabrielbbaldez:stacktale-mcp:1.0.0`, so the Claude Code plugin keeps installing the old server after a release.
- `.claude-plugin/marketplace.json` — pins `"version": "1.0.0"`.
- `plugins/stacktale/.claude-plugin/plugin.json` — same.

This is not hypothetical: two instances of exactly this drift were already found and fixed by hand today (a README dependency pinned at a version that is on no repository, and the action's README pinning a tag that does not exist).

## What to do

Either extend the version-bump step in `.github/workflows/prepare-release.yml` to `sed` these files too, **or** — better, because it catches drift from any source rather than just from the release path — add a small CI check that fails when they disagree with the project version.

The check is a few lines of shell: read `<version>` from the parent `pom.xml`, grep the pinned versions out of the three JSON files, compare. Run it in the existing CI workflow.

If you take the check route, note that during development the parent is a `-SNAPSHOT` while the plugin should point at the last *released* version — so the rule is not plain equality. Decide the rule, state it in a comment, and make the failure message say what to change.

## Files

- `.github/workflows/prepare-release.yml`
- `plugins/stacktale/.mcp.json`, `.claude-plugin/marketplace.json`, `plugins/stacktale/.claude-plugin/plugin.json`
- wherever you put the check (`.github/workflows/ci.yml`, or a script under `scripts/` beside `check-readme-compat.sh`, which is the existing precedent for this kind of guard)

## How to verify

Change one pinned version to something wrong and confirm CI goes red with a message that names the file. Then run the release workflow's bump step locally with a dummy version and confirm every pinned file moves together.

PR #160, opened 2026-08-03
Add standalone examples for JUL, Spring Boot MVC, and WebFlux

asmitayush3021:
## Summary

This PR adds standalone examples demonstrating Stacktale with:

- Plain Java (JUL)
- Spring Boot MVC
- Spring Boot WebFlux

## Verification

- ✅ Plain Java example runs successfully.
- ✅ Spring Boot MVC generates AI-oriented error reports.
- ✅ Spring Boot WebFlux generates AI-oriented error reports, including errors across Reactor scheduler hops.

## Notes

The WebFlux example was tested against my locally built `1.1.0-SNAPSHOT`.

When using the published `0.4.0` artifact from Maven Central, the application fails to start because of the known servlet auto-configuration issue discussed in #62. With the current `main` branch build, the example runs successfully.

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -1,0 +1,36 @@
+name: Build Examples
+
+on:
+  pull_request:
+    paths:
+      - "examples/**"
+      - ".github/workflows/examples.yml"
+  push:
+    paths:
+      - "examples/**"
+      - ".github/workflows/examples.yml"
+
+jobs:
+  build-examples:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        example:
+          - examples/plain-java-jul
+          - examples/spring-boot-mvc
+          - examples/spring-boot-webflux
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up JDK 17
+        uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: 17
+          cache: maven
+
+      - name: Compile example
+        working-directory: ${{ matrix.example }}
+        run: mvn -B -q compile
--- a/examples/plain-java-jul/README.md
+++ b/examples/plain-java-jul/README.md
@@ -1,0 +1,51 @@
+# Plain Java JUL (java.util.logging) Example
+
+A minimal, standalone plain Java project demonstrating direct `java.util.logging` (JUL) integration using `StacktaleJulHandler`.
+
+## What This Example Demonstrates
+
+- **Direct JUL Integration**: Uses the JDK's built-in `java.util.logging` framework directly via `StacktaleJulHandler` without requiring Spring Boot or Logback.
+- **Declarative `logging.properties`**: Demonstrates configuring the handler, package root markers, and output file declaratively using standard JUL `logging.properties`.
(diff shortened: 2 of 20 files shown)


GabrielBBaldez (changes requested):
Three runnable projects, each with its own README, and the WebFlux one covers the scheduler-hop case that is the hardest thing to demonstrate. This is #62 delivered.

One thing blocks it, and it is the kind that only shows up from outside.

### The three examples pin three different versions, and none is the released one

| example | pinned |
|---|---|
| `plain-java-jul` | `0.5.0` (via `stacktale.version`) |
| `spring-boot-mvc` | `0.4.0` |
| `spring-boot-webflux` | `1.1.0-SNAPSHOT` |

Current release is **1.1.0**.

The SNAPSHOT is the one that actually breaks. It resolves nowhere:

```
Central 1.1.0:           HTTP 200
Central 1.1.0-SNAPSHOT:  HTTP 404
```

All three compiled when I ran them, including WebFlux — because my local `~/.m2` has `1.1.0-SNAPSHOT` from building this repo. Yours almost certainly does too. Anyone who clones the examples without having built stacktale from source gets a resolution failure, and that is precisely the person an example exists for.

`0.4.0` and `0.5.0` are worth noticing too: neither is wrong in the sense of being made up — each was "current" at some past moment. That is the shape of the real problem below.

### Nothing builds them, so they will drift again

`examples/` is not in the reactor and CI reports `no checks reported` on this branch. Nothing compiles these projects, so the next release moves the library and the examples stay where they are, silently. The version spread already in this PR is that failure having happened once.

Worth fixing here rather than later, and it is small:

1. Pin `1.1.0` in all three. Keep the `stacktale.version` property idiom from the JUL example and use it in the other two — one place per project.
2. Add a CI job that builds each example directory. They should stay **out** of the reactor: a reactor build would resolve them against the in-tree SNAPSHOT, which is the opposite of what an example must prove. A loop over `examples/*/pom.xml` running `mvn -B -q compile` is enough.
3. If you want the guard to be complete, have that job assert the pinned version matches the latest release rather than only that it resolves — otherwise it stays green on `1.1.0` forever after `1.2.0` ships.

Point 3 is optional; 1 and 2 are what I would like before merging.

Nothing else in the diff needs changing — the JUL `logging.properties`, the exception with real state on it, and the WebFlux controller all read the way I would want someone meeting the project to see it.

GabrielBBaldez (approved):
Both points, and the second one done properly — a matrix job per example rather than one step that stops at the first failure, so a break in the JUL example still tells you whether WebFlux is fine.

Verified rather than assumed. The three now build against a **completely empty local repository**:

```
mvn -B -q compile -Dmaven.repo.local=<empty dir>
  plain-java-jul        OK
  spring-boot-mvc       OK
  spring-boot-webflux   OK
```

That is the check that matters here. Everything resolved from Central, which is what proves `1.1.0` works for someone who has never built this project — the exact case `1.1.0-SNAPSHOT` failed and that my own `~/.m2` had been hiding.

`Build Examples` is green on all three legs of the matrix. It needed approving first — fork PRs sit behind the workflow-approval gate here, which is why the run showed as `no checks reported` rather than as anything you did.

I left the third point out of the merge on purpose. It was marked optional and it is a different problem: the pin is right today and goes stale the moment 1.2.0 ships, which is the same shape as #142 for the plugin and marketplace versions. Opening it separately rather than growing this PR.

Merging. Thanks for coming back to it this quickly — and for the WebFlux example specifically, since the scheduler-hop case is the one that is hardest to show and easiest to get wrong.

PR #164, opened 2026-08-04
Add JUnit Platform 1.10 compatibility floor

### LINKED ISSUE #155: stacktale-junit has no stated or tested JUnit floor
`stacktale-junit` compiles against `junit-platform-launcher` at `${junit.version}` — currently **6.1.2**, the JUnit 6 line. The Compatibility table in the README has rows for Java, Logback, Log4j2 and Spring Boot, and none for JUnit. `compat.yml` has matrix entries for the same four and none for JUnit either.

Most projects that would install this are still on JUnit 5.x, where the launcher is 1.x.

The listener only touches API that has been in the Platform since 1.x — `TestExecutionListener`, `TestIdentifier`, `TestPlan`, `MethodSource` / `ClassSource`, `TestTag` — so it very likely works. "Very likely" is the problem: this is the one module whose entire job is to stop an agent seeing a green result on a red build. If it silently fails to load on a 5.x project, the failure mode is exactly the one it exists to prevent, and nobody finds out.

## What to do

1. Pick and state a floor. `junit-platform-launcher` 1.10 (JUnit 5.10) is a reasonable line — old enough to cover the field, new enough to not contort the code.
2. Add a matrix entry to `.github/workflows/compat.yml` pinning it, alongside the Logback and Spring Boot floors:
   ```yaml
   - name: "JUnit Platform 1.10 (floor)"
     args: "-pl stacktale-core,stacktale-junit -am -Djunit.version=..."
   ```
   Note the launcher and Jupiter version properties are separate lines in JUnit 5 (`5.10.x` for Jupiter, `1.10.x` for the Platform) but a single line in JUnit 6 — so this may need its own property rather than reusing `${junit.version}`. Work out the shape and say what you chose.
3. Add the row to the README's Compatibility table. `scripts/check-readme-compat.sh` guards the other four rows; extending it to this one is optional but in keeping.

## Verify

The new matrix job passes on the floor, and `mvn -pl stacktale-junit -am test -D<the floor>` passes locally. If it does not, that is the more valuable outcome — it means the module has a floor higher than anyone documented.

Sarthak-Vatsa:
## What & why

Adds JUnit Platform 1.10 as the compatibility floor for stacktale-junit, with separate Platform and Jupiter version properties so CI can verify the JUnit 5.10 runtime while preserving JUnit 6.1.2 as the default.

Adds the floor to the compatibility matrix and documents it in the README compatibility table.

Closes #155

## How it was verified

## Checklist

- [x] The issue was claimed with a comment before starting (see
      [CONTRIBUTING](https://github.com/stacktale/stacktale/blob/main/CONTRIBUTING.md#claim-the-issue-before-you-start)
      — didn't claim it? Open the PR anyway, just say so)
- [x] One logical change (unrelated fixes belong in their own PR)
- [ ] Behavior changes arrive with the test that demanded them; bug fixes start from a
      failing test
- [x] **No golden-file test changed** — `st/1` is a public API. If a golden test did
      change, that is a deliberate format change: say so below, and update
      [docs/FORMAT.md](https://github.com/stacktale/stacktale/blob/main/docs/FORMAT.md)
- [ ] New config property? It has a setter (Joran naming), a default that preserves
      current behavior, coverage in `LogbackXmlConfigTest`, and a row in the README
      config table

## AI assistance (optional)

     
AI assistance was used to understand the JUnit Platform/Jupiter versioning and reason through the compatibility-testing approach.

### DIFF
--- a/.github/workflows/compat.yml
+++ b/.github/workflows/compat.yml
@@ -31,6 +31,8 @@ jobs:
             args: "-Dspring-boot.version=3.3.13"
           - name: "Spring Boot 3.5"
             args: "-Dspring-boot.version=3.5.6"
+          - name: "JUnit Platform 1.10 (floor)"
+            args: "-pl stacktale-core,stacktale-junit -am -Djunit.platform.version=1.10.0 -Djunit.jupiter.version=5.10.0"
     name: ${{ matrix.name }}
     steps:
       - uses: actions/checkout@v7
--- a/README.md
+++ b/README.md
@@ -672,6 +672,7 @@ each backed by a passing build:
 | Logback | 1.4+ | 1.6.x |
 | Log4j2 | 2.20+ | 2.26.x |
 | Spring Boot *(starter)* | 3.2+ | 3.5.x |
+| JUnit Platform *(stacktale-junit)* | 1.10+ | 6.1.x |
 
 The Spring Boot starter follows Boot's own Logback version (1.5 on Boot 3.4+); the
 `stacktale` and `stacktale-log4j2` artifacts work down to Logback 1.4 on their own.
--- a/pom.xml
+++ b/pom.xml
@@ -53,6 +53,8 @@
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
     <logback.version>1.6.1</logback.version>
     <junit.version>6.1.2</junit.version>
+    <junit.platform.version>${junit.version}</junit.platform.version>
+    <junit.jupiter.version>${junit.version}</junit.jupiter.version>
     <assertj.version>3.27.7</assertj.version>
     <spring-boot.version>3.5.16</spring-boot.version>
     <log4j2.version>2.26.1</log4j2.version>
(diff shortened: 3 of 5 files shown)


GabrielBBaldez (approved):
Exactly the shape the issue was uncertain about. Separate `junit.platform.version` and `junit.jupiter.version`, both defaulting to `${junit.version}` so nothing changes for anyone not passing the flags — that is the part I could not name in advance, and it is right.

Verified the properties actually reach the classpath rather than take the green build as proof:

```
              default                       with the floor flags
launcher      6.1.2                         1.10.0
junit-jupiter 6.1.2                         5.10.0
```

Worth spelling out why that check mattered here. A matrix entry passing a property nothing reads produces a **passing job that tested the default twice** — the same failure mode as the Plugin Verifier trap in the sibling repo, where a compat leg looked green while exercising nothing. `dependency:tree` is what separates the two, and it comes back with the right answer.

The suite then passes on the floor: 8 tests in `stacktale-junit` against Platform 1.10 with Jupiter 5.10. So the listener really does only use API that has been there since 1.x, which until now was an assumption in a comment.

Extending `check-readme-compat.sh` was not required and is the best part of this. I broke the new row deliberately to confirm it is not decoration:

```
README compatibility table: JUnit Platform says 9.9.x but the build tests up to 6.1.2
```

`mvn verify` green across all eight modules here, CI green on the PR. Merging.

Thanks @Sarthak-Vatsa — the property split was the open question in the issue, and extending the README guard wasn't asked for.

PR #167, opened 2026-08-07
fix(report): summarize st-json/1 logs in the report action

### LINKED ISSUE #141: The report action silently finds nothing in an st-json/1 log

syf2211:
## Summary

The report composite action now recognizes `st-json/1` NDJSON logs (`format=json`) and summarizes `report` entries instead of silently reporting zero results.

## Motivation

Fixes #141. When a project writes `errors-ai.log` as NDJSON, `summarize.sh` previously grepped for text-format `━━━ END #` delimiters that never exist in JSON output. A red build therefore looked identical to a green one: no PR comment, no job summary, and `count=0`.

## Changes

- Detect `st-json/1` from the first non-blank header line (with trim + substring fallback when `jq` cannot validate)
- Count and render `type=report` lines with `jq` when available
- When `jq` is absent or unusable, write a clear step-summary notice and keep artifact upload working
- Show the step summary whenever `summarize.sh` writes output, not only when `count > 0`
- Upload artifacts whenever `upload-artifact` is enabled (not gated on `count`)
- Add fixture + workflow self-tests (whitespace header, no-`jq` fallback, composite action e2e)

## Tests

- `bash .github/actions/report/summarize.sh .github/actions/report/fixtures/sample-st-json.ndjson 1 /tmp/out.md` → count `2`
- Text-format golden self-test path unchanged (still expects 3 complete reports)
- Workflow cases added in `.github/workflows/report-action.yml` for st-json/1

## Notes

- PR comments are still posted only when `count > 0` (unchanged); the jq-absent notice appears in the job summary only.
- Repeat/session/storm JSON lines are intentionally excluded from the inline summary, matching how the MCP server treats them.

Fixes #141

### DIFF
--- a/.github/actions/report/README.md
+++ b/.github/actions/report/README.md
@@ -42,7 +42,10 @@ Useful for gating later steps.
   line has not arrived rather than show half a report.
 - **A missing log is not a failure.** A green build has nothing to say, and the action
   exits quietly with `count=0`.
-- **No runtime dependency.** The extractor is awk, so it works on any runner.
+- **No runtime dependency for text logs.** The st/1 extractor is awk, so it works on any
+  runner. **st-json/1** logs (`format=json`) are summarized with `jq` when it is on the
+  PATH (GitHub-hosted runners include it). Without `jq`, the action prints a clear notice
+  and still uploads the log as an artifact.
 
 ## Where the reports come from
 
--- a/.github/actions/report/README.md
+++ b/.github/actions/report/README.md
@@ -52,3 +52,7 @@ Useful for gating later steps.
 The library writes them. A failing *test* only produces one if `stacktale-junit` is on the
 test classpath — without it, a red `mvn test` writes nothing and this action has nothing to
 show. See the [quickstart](../../../README.md#quickstart).
+
+Projects configured with `format=json` write **st-json/1** NDJSON (one JSON object per line)
+instead of the default **st/1** text blocks. The action summarizes both when `jq` is
+available.
--- a/.github/actions/report/action.yml
+++ b/.github/actions/report/action.yml
@@ -43,11 +43,11 @@ runs:
         fi
         count=$(bash "${{ github.action_path }}/summarize.sh" "$log" "${{ inputs.max-reports }}" "$RUNNER_TEMP/stacktale-summary.md")
         echo "count=$count" >> "$GITHUB_OUTPUT"
-        if [ "$count" -gt 0 ]; then
+        if [ -s "$RUNNER_TEMP/stacktale-summary.md" ]; then
           cat "$RUNNER_TEMP/stacktale-summary.md" >> "$GITHUB_STEP_SUMMARY"
         fi
 
-    - if: steps.summarize.outputs.count != '0' && inputs.upload-artifact == 'true'
+    - if: inputs.upload-artifact == 'true'
       uses: actions/upload-artifact@v4
       with:
         name: stacktale-errors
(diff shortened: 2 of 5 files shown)


GabrielBBaldez (changes requested):
The implementation is right and the `st/1` path is untouched — I checked both rather than trust the green half of CI. One line in the new test contradicts itself, and that is the whole failure.

## The failing case asserts something it told the script not to produce

`.github/workflows/report-action.yml:81` calls:

```bash
count=$(bash .github/actions/report/summarize.sh errors-ai.log 1 /tmp/st-json-summary.md)
```

`max=1`, so the summary renders the first report and says *"Showing the first 1."* Line 88 then asserts:

```bash
grep -q 'ERROR (no exception): checkout timed out' /tmp/st-json-summary.md
```

That is the **second** report in the fixture — the one `max=1` deliberately left out. `count` is 2 and the count check passes, which is why the failure lands three assertions later and reads like a rendering bug rather than an argument.

Changing the `1` to a `2` fixes it, and is the better fix than deleting the assertion: the `noException` headline branch is worth covering, and that fixture entry exists to cover it.

I ran the five assertions from that step both ways:

| | `max=1` | `max=2` |
|---|---|---|
| count is 2 | ok | ok |
| `2 error reports` | ok | ok |
| `IllegalStateException: payment gateway refused` | ok | ok |
| `ERROR (no exception): checkout timed out` | **fails** | ok |
| `json0001` | ok | ok |
| `repeat` entries stay out | ok | ok |

## The rest holds up

`summarize.sh` on the new fixture returns `2` with the headline, id and timestamp rendered as the description says, and `repeat` entries correctly excluded from the count.

The text format did not regress — the same golden self-test the action already had:

```
count=3          (the truncated block still excluded)
Showing the first 2   ok
3 error reports       ok
dead9999 leaked       no
```

Detecting the header from the first *non-blank* line with a trim, and keeping a substring fallback for when `jq` cannot validate it, is more careful than the issue asked for. So is writing a notice when `jq` is absent instead of returning a silent zero — that failure mode is the one #141 was filed about, and reintroducing it in the fallback would have been easy to miss.

Note I could not run `jq` locally, so I stood up a small stand-in covering the six invocations the script makes. The numbers above come from that; the CI run is the authority once the `1` becomes a `2`.

GabrielBBaldez (approved):
Pushed the one-character fix rather than send it back for it — `max=1` to `max=2` on the st-json case, so the second fixture entry is rendered and the `noException` assertion has something to find. `Report action` goes from failure to success on that alone.

Also merged `main` in, because #168 landed on the same two files while this was open and left the PR unmergeable. One real conflict, in `action.yml`: main bumped `upload-artifact` to v7, this branch removed the `count != '0'` gate. Kept both — the gate stays off, since an st-json/1 log on a runner without `jq` summarizes to 0 and that is precisely when the raw file is worth attaching. Taking main's side wholesale would have quietly undone half of what this PR is for.

Re-checked both formats after the merge: st-json `count=2`, text `count=3` with the truncated block still excluded.

Thanks @syf2211 - the header detection reading the first non-blank line with a trim, and the honest notice when `jq` is missing, are both more careful than the issue asked for. That notice matters most: returning a silent zero there is the exact failure #141 was about, and reintroducing it in the fallback would have been easy to do.

Issue #40, opened 2026-07-10
Supply chain: SBOM (CycloneDX) and build provenance on releases


GabrielBBaldez:
﻿Modern OSS hygiene for a library people put INSIDE their apps: publish a CycloneDX SBOM per module and attach SLSA-style build provenance (GitHub's artifact attestations) to release artifacts.

Done when: releases carry SBOM files and provenance attestations, documented in the README security note.

---

## Guidance for a first-time contributor

Mostly configuration â€” two plugins and a workflow step. No production code changes.

**1. SBOM per module.** Add the CycloneDX Maven plugin to the root `pom.xml` build section
so every module inherits it:

```xml
<plugin>
  <groupId>org.cyclonedx</groupId>
  <artifactId>cyclonedx-maven-plugin</artifactId>
  <executions>
    <execution>
      <phase>package</phase>
      <goals><goal>makeAggregateBom</goal></goals>
    </execution>
  </executions>
</plugin>
```

Pin the version like every other plugin here (a `<properties>` entry, matching local style).
Check the output lands in `target/` for each module and that `mvn -B verify` still passes.

**2. Attach it to releases.** The publishing flow lives in
[`.github/workflows/release.yml`](../blob/main/.github/workflows/release.yml). Add the
generated SBOM files to the release artifacts there. Read that workflow before editing â€”
it also signs artifacts, and the ordering of steps matters.

**3. Build provenance.** Add GitHub's `actions/attest-build-provenance` step after the
artifacts are built. It requires `permissions: id-token: write` and `attestations: write`
on the job â€” the workflow will fail confusingly without them, so set them explicitly.

**4. Document it.** One short paragraph in the *Supply chain* part of `SECURITY.md`, which
already covers the dependency surface and release signing. That is the natural home; do not
open a new README section for it.

**Careful:** do not change what gets published to Maven Central, and do not touch the GPG
signing steps. If you cannot test the release workflow (you can't â€” it needs the repo's
secrets), say so in the PR; CI will validate the build side, and the maintainer validates the
release side on the next tag.

PR #175, opened 2026-08-10
infra: add CodeQL and dependency-review workflows

### LINKED ISSUE #156: infra: add CodeQL and dependency-review
Nothing in `.github/workflows/` runs static analysis or checks dependencies on a PR. For most libraries that would be a nice-to-have; for this one the attack surface is unusually specific:

- `FieldExtractor` reflects over **user** exception classes and calls `trySetAccessible`
- `Redactor` compiles **user-supplied** regexes and runs them on the application's logging thread — that is a live open issue (#118, ReDoS)
- `stacktale-agent` shades ByteBuddy and instruments user bytecode

## What to do

Two template workflows, both small:

- **CodeQL** (`github/codeql-action`) on the `java` language, on push to main and on PRs. It gives Security tab alerts and feeds the OpenSSF Scorecard.
- **Dependency review** (`actions/dependency-review-action`) on PRs, which blocks a dependency bump that introduces a known vulnerability — useful given how many dependabot PRs this repo takes.

CodeQL needs `security-events: write`, declared in the workflow. That is available now — the org setting that forced workflow tokens to read-only has been lifted.

## Scope

This is not #40. That issue is about SBOM and build provenance — proving what the artifact *is*. This is about finding problems in the source before they ship. Adjacent, not overlapping.

## Verify

The Security tab shows a completed CodeQL scan, and a PR bumping a dependency with a known CVE gets blocked by dependency-review.

dchaudhari7177:
Closes #156.

Two small workflows, as the issue describes.

## CodeQL (`.github/workflows/codeql.yml`)

Runs on push to `main`, on PRs, and weekly. The weekly run is the one that matters for a library this size — without it, a newly published query only reaches code somebody happens to touch. Scheduled at 05:00 Monday, an hour ahead of `compat.yml` so the two don't land together.

**Build mode is `manual`, not `autobuild`.** Autobuild runs the full reactor including tests, duplicating what CI already does on every PR; the analysis only needs class files, so the step is `mvn -B -DskipTests compile`. It has to sit between `init` and `analyze` — manual mode analyses whatever that build traces, and an empty trace reports success having examined nothing, which is the quiet failure mode worth avoiding here.

`security-events: write` is declared on the job, as the issue notes is now possible. Without it the scan completes and only the upload fails, which reads as flakiness rather than a missing permission.

## Dependency review (`.github/workflows/dependency-review.yml`)

`pull_request` only — the action diffs the base's dependency graph against the head's, so a push to `main` gives it nothing to compare.

Two judgement calls, both written into the file:

- **`fail-on-severity: moderate`**, not `low`. Given how many dependabot PRs this repo takes, `low` would hold them up on advisories against code paths the library never reaches. Lower findings still appear in the job summary.
- **`comment-summary-in-pr` deliberately unset.** It needs `pull-requests: write`, and a PR from a fork gets a read-only token — so it would fail on exactly the external contributions this check is most useful for. The findings are in the job summary either way.

## Verification

Both files parse as YAML. Action majors are current as of today — `codeql-action@v4`, `dependency-review-action@v5`, `checkout@v7`, `setup-java@v5` — matching the pins already used across the other workflows, and `java-version: 21` / `distribution: temurin` / `cache: maven` follow `ci.yml`.

I could not run `mvn` locally (not installed on this machine), so the compile step is unexercised here; it is the same reactor `ci.yml` already builds with `mvn -B verify`. The real check is the issue's own: the Security tab showing a completed scan, and a CVE-bearing bump getting blocked — both of which need this merged to `main` first.

## Scope

Not #40 — that is SBOM and build provenance, about proving what the artifact *is*. This is finding problems in the source before it ships. Adjacent, not overlapping.

---

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,53 @@
+name: CodeQL
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+  schedule:
+    # Weekly, so a newly published query finds existing code rather than waiting
+    # for someone to touch the file it applies to. Offset from compat.yml's 06:00.
+    - cron: "0 5 * * 1"
+
+permissions:
+  contents: read
+
+jobs:
+  analyze:
+    name: Analyze Java
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      # The upload at the end of `analyze` goes through the code-scanning API.
+      # Without this the scan itself succeeds and only the upload fails, which
+      # reads as a flaky job rather than a missing permission.
+      security-events: write
+      # Read-only, and only so CodeQL can look up the workflow run it belongs to.
+      actions: read
+    steps:
+      - uses: actions/checkout@v7
+
+      - uses: actions/setup-java@v5
+        with:
+          distribution: temurin
+          java-version: 21
+          cache: maven
+
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v4
+        with:
+          languages: java-kotlin
+          build-mode: manual
+
+      # `autobuild` would run the full reactor, tests included, duplicating what
+      # CI already does on every PR. The analysis only needs class files, so this
+      # compiles and stops. It must sit between init and analyze: manual mode
+      # traces this build, and an empty trace analyses nothing while still
+      # reporting success.
+      - name: Build classes for analysis
+        run: mvn -B -DskipTests compile
+
+      - name: Perform CodeQL analysis
+        uses: github/codeql-action/analyze@v4
+        with:
+          category: "/language:java-kotlin"
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -1,0 +1,27 @@
+name: Dependency review
+
+# Pull requests only: the action diffs the dependency graph of the base against
+# the head, so it has nothing to compare on a push to main.
+on: pull_request
+
+permissions:
+  contents: read
+
+jobs:
+  review:
+    name: Review dependency changes
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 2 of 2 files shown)


GabrielBBaldez (approved):
The code is right and CodeQL is green — `Analyze Java` and `CodeQL` both pass on this PR. `Review dependency changes` fails for a repository setting, not for anything in the diff:

```
Dependency review is not supported on this repository.
Please ensure that Dependency graph is enabled
```

I tried to enable it over the API; it is not a writable `security_and_analysis` field, so it needs the toggle at **Settings → Code security → Dependency graph**. Holding the merge until then rather than landing a check that is red for a reason nobody reading the PR caused.

Three things in here are better than the issue asked for.

**`build-mode: manual` with `mvn -B -DskipTests compile`.** `autobuild` would run the full reactor with tests, duplicating CI on every PR. And the comment names the real hazard:

> an empty trace analyses nothing while still reporting success

That is the failure mode this repo keeps finding — a green run that checked nothing — and it is called out before it happens rather than after.

**No `comment-summary-in-pr`, with the reason.** A fork PR gets a read-only token, so the comment would fail on exactly the contributions worth checking. That trap cost this project a day earlier this week on a different workflow; you avoided it from the start.

**`fail-on-severity: moderate` rather than `low`**, so a Dependabot PR does not stall on an advisory against a path the library never reaches, with everything below still visible in the summary. That is the trade that decides whether a gate survives its first month.

Thanks @dchaudhari7177 - the fork-token note especially; that one is usually learned the expensive way.

Merging as soon as the toggle is on.

GabrielBBaldez:
Dependency graph is on now, and `Review dependency changes` goes from failure to success on a re-run — the whole check set is green. Merging.

Issue #132, opened 2026-07-25
infra: add a link checker to CI so README links can't rot again


GabrielBBaldez:
A relative link in `README.md` pointed at `stacktale/src/main/java/.../StacktaleExecutors.java` for several releases after the file moved to `stacktale-core/` in the multi-module split. It was fixed by hand in 6e55ee5, but nothing stops the next one — and a 404 from the README lands exactly when a reader is trying to go deeper.

## What to do

Add a link-check job to `.github/workflows/ci.yml` (or its own scheduled workflow):

```yaml
- name: Check links
  uses: lycheeverse/lychee-action@v2
  with:
    args: --no-progress --exclude-mail README.md CONTRIBUTING.md SECURITY.md 'docs/**/*.md'
    fail: true
```

Notes:
- Run it on `pull_request` and on a weekly `schedule` — external links rot on their own timetable, and you don't want a third-party outage failing an unrelated PR. Consider `fail: true` only for the scheduled/main run.
- Add a `.lycheeignore` for anything intentionally unreachable (e.g. `javadoc.io` rate-limits and can 5xx).
- `docs/site/index.html` is worth including too.

## Verify

Break a link on a branch and watch the job go red; the scheduled run stays green on main.

sagarbarnwal722-dev:
Hi! I'd like to work on this issue as my first open-source contribution.Could you please assign it to me? Thank you!

GabrielBBaldez:
Feel free to take it. Thanks for your contribution!

If you like the project and find it useful, I'd really appreciate it if you could leave it a ⭐ on GitHub. It helps the project gain visibility and motivates me to keep improving it.

Thanks again!

sagarbarnwal722-dev:
> Feel free to take it. Thanks for your contribution!
> 
> If you like the project and find it useful, I'd really appreciate it if you could leave it a ⭐ on GitHub. It helps the project gain visibility and motivates me to keep improving it.
> 
> Thanks again!

Hi! I'd like to work on this. Do you prefer adding the Lychee link checker as a new job in the existing ci.yml, or as a separate scheduled workflow (e.g. link-check.yml)? I'll follow whichever approach you prefer.

GabrielBBaldez:
Separate workflow, and the reason is worth having in mind because it decides the trigger too.

[...]

PR #176, opened 2026-08-10
infra: fail when the examples stop pinning the current release

### LINKED ISSUE #161: infra: nothing notices when the examples' pinned version goes stale
`examples/` pins `1.1.0` in three `pom.xml` files. That is correct today and wrong the day 1.2.0 ships — nothing connects the two, so the examples will quietly demonstrate an old version until someone notices.

This already happened once, before the folder was even merged. The first revision of #160 carried `0.5.0`, `0.4.0` and `1.1.0-SNAPSHOT` across the three projects; each had been "current" at some past moment, and one of them resolved nowhere at all.

`Build Examples` catches an example that stops *compiling*. It cannot catch one that compiles perfectly against a version nobody should be using any more.

## What to do

Add a step to `.github/workflows/examples.yml` that compares the pinned version against the latest release and fails when they differ.

The latest release is available without an API token:

```bash
latest=$(curl -fsSL https://api.github.com/repos/stacktale/stacktale/releases/latest | jq -r .tag_name)
latest=${latest#v}
```

Then read what each example pins. Two of them declare it inline and one through a `stacktale.version` property, so the robust way is to ask Maven rather than to grep:

```bash
mvn -q -f "$example/pom.xml" help:evaluate -Dexpression=stacktale.version -DforceStdout
```

which means the two inline ones should move to the same property first — worth doing anyway, since it puts the version in one place per project.

Two judgement calls to make and write down:

- **What to do about a release that has not happened yet.** Right after a version bump on `main` and before the tag is pushed, the examples legitimately point at the older released version. Failing there would make the release process fight this check. Warning rather than failing during that window, or comparing against the tag rather than the branch, are both defensible — pick one and say why.
- **Where the failure should land.** A red X on a PR that only touched a README is confusing when the cause is a release that happened last week. A scheduled run that opens an issue may serve better than a PR check; #132 makes the same trade for link rot.

## Verify

Bump one example's pin to a version below the latest release and watch the check object, naming the example and both versions. Put it back and watch it pass.

Related: #142 is this same problem for the plugin and marketplace versions.

dchaudhari7177:
Closes #161.

## The drift has already happened

Before adding any check: **the latest release is `v1.2.0`, and all three examples pinned `1.1.0`.** They have been demonstrating a version one minor behind. Bumped to `1.2.0`, with `stacktale-jul` and `stacktale-spring-boot-starter` both confirmed present on Maven Central at that version (HTTP 200 on the `.pom`).

Without this, the new check would land red on merge — which is the check working, but not a useful first impression.

## Normalising the pins

The two inline pins move to the same `stacktale.version` property `plain-java-jul` already used. That is what lets the check ask Maven for one expression instead of knowing which of two shapes each project uses, and it puts the literal in one place per project — worth doing on its own, as the issue notes.

## The two judgement calls

**Where the failure lands.** A new `version-drift` job in `examples.yml`, triggered `schedule` *as well as* the existing paths. The cause of this drift is a **release**, not a commit under `examples/` — so the existing path filters would never fire for it, and the check would sit silent until someone happened to touch an example. That is precisely the failure mode the issue describes. Weekly at 07:00 Monday, after `compat.yml` (06:00) and `codeql.yml` (05:00).

It does **not** open an issue. That needs `issues: write` plus dedupe logic to avoid filing the same one every week — more machinery than the signal warrants, when a red scheduled run is already visible in Actions. Happy to switch to the #132 approach if you'd rather they be consistent.

**A release that has not happened yet.** Compared with `sort -V`, treating the two sides differently:

| pinned vs latest | outcome | why |
|---|---|---|
| equal | pass | |
| **older** | **fail** | this is the bug |
| newer | warn | between a version bump on `main` and the tag being pushed, the examples legitimately point ahead — failing here would make the release process fight this check |

## Verification

Dry-run against the live repository (`latest = v1.2.0`):

```
  1.1.0 vs 1.2.0 -> FAIL (behind)
  1.0.0 vs 1.2.0 -> FAIL (behind)
  9.9.9 vs 1.2.0 -> warn (ahead)
```

The workflow parses as YAML and both jobs are present.

**Not tested:** Maven is not installed in my environment, so the `help:evaluate` step itself is unexercised — CI is the first real run of it. The issue's own verification (bump a pin below the latest and watch it object, then put it back) needs this on `main` first.

---

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -9,6 +9,15 @@ on:
     paths:
       - "examples/**"
       - ".github/workflows/examples.yml"
+  schedule:
+    # The drift this catches is caused by a *release*, not by a commit under
+    # examples/, so the path filters above would never fire for it. Weekly,
+    # Monday 07:00 UTC — after compat.yml (06:00) and codeql.yml (05:00).
+    - cron: "0 7 * * 1"
+  workflow_dispatch:
+
+permissions:
+  contents: read
 
 jobs:
   build-examples:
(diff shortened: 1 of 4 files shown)


GabrielBBaldez (approved):
Both open questions in the issue answered, and answered better than the options I listed.

**When the release has not happened yet.** I offered "warn instead of fail during that window" or "compare against the tag". A weekly `schedule` sidesteps the question entirely — the drift is caused by a *release*, not by a commit, so a job that only runs on `examples/**` would never have fired for it in the first place. The comment says exactly that, which is the part that will stop someone "simplifying" it back into the path-filtered job later.

**Where the failure lands.** On a scheduled run, not on a PR that touched a README. That was the concern; a red X whose cause is a release from last week teaches people to ignore red Xs.

Asking Maven for the property instead of grepping the pom is the right call now that the version is behind an indirection — and moving the two inline pins onto `stacktale.version` is what makes that possible. One literal per project.

Verified rather than read. All three resolve `1.2.0` against the current release, and dropping one to `1.1.0` produces:

```
DRIFT: spring-boot-mvc pina 1.1.0, release atual é 1.2.0
```

`Examples pin the current release` is green on the PR alongside the three matrix legs.

Thanks @dchaudhari7177 - the schedule is the insight here; both options I wrote down were worse.

PR #177, opened 2026-08-10
infra: guard the docs' example reports against shapes the renderer cannot emit

### LINKED ISSUE #154: infra: nothing checks that the example reports in the docs are producible
Three separate doc bugs found in one audit turned out to be the same bug:

- the console line in the README printed a **relative** path, months after the code started printing an absolute one
- the `stacktale-junit` example was missing an MDC key the listener always writes, and stopped before the stack, `env:` and the closing delimiter, while being introduced as a whole report
- `… 39 collapsed (spring ×24, tomcat ×11)` — 24 + 11 = 35. `StackDistiller` counts every frame of a run into a group, so the parts always sum to the total. The landing page had its own impossible version of the same line.

All three are docs claiming output the renderer cannot produce. Nothing checks that, even though the authoritative text already exists in `stacktale-core/src/test/resources/golden/`.

## What to do

A test that extracts the fenced blocks from `README.md` (and ideally `docs/site/index.html`) and asserts they are consistent with the real format. Full equality against a golden file is too strict — the docs legitimately use a different app name and shorter stacks — so check the properties that were actually wrong:

- a block that opens with `━━━ ERROR #<id> ━━━` also closes with `━━━ END #<id> ━━━`, same id
- every `… N collapsed (a ×x, b ×y)` satisfies `x + y + … == N`
- section headers match the real shape: `story (…, last N events, …)`, `stack (distilled, N of M frames)`
- if a block has a `story`, it also has `env:` — i.e. it is not truncated mid-report

`FORMAT.md` §3 is the reference for what a block must contain, and `scripts/check-readme-compat.sh` is the precedent for this kind of guard: a script the CI runs against the README, with a failure message that says what to change.

Put it wherever fits — a script beside the compat one, or a test in `stacktale-core` that reads the README from the repo root.

## Verify

Reintroduce `… 39 collapsed (spring ×24, tomcat ×11)` and watch it go red with a message naming the line.

dchaudhari7177:
Closes #154.

`scripts/check-readme-reports.sh`, run from `ci.yml` beside the existing compat and plugin-manifest guards.

Full equality against `stacktale-core/src/test/resources/golden/` would be too strict — the docs use a different app name and shorter stacks on purpose — so it checks the properties that were actually wrong, per FORMAT.md §3:

1. a block opening `━━━ ERROR #<id>` closes with `━━━ END #<id>`, same id
2. every `… N collapsed (a ×x, b ×y)` satisfies `x + y + … == N`
3. `story (…, last N events, …)` and `stack (distilled, N of M frames)` keep their real shape — plus `N <= M`, since a stack cannot distil to more frames than it started with
4. a block with a `story` also has an `env:` line, so a truncated report cannot be introduced as a whole one

## Two decisions worth flagging

**Rule 2 is checked everywhere, not only inside fenced blocks.** Two of the three original bugs were in running prose — a table row and the landing page — not in an example block. `docs/site/index.html` is checked too when present.

**Counts are read as the last number in each comma-separated part**, rather than by matching the `×`. That sign is multibyte and `mawk` — the default `awk` on the runner — is byte-oriented, so matching it is not portable. Reading the last number also survives a framework whose name contains a digit (`log4j2 ×5`), which "first number in the part" would not.

## Verification

Passes against the current docs, so this lands green.

Each rule was then proved by injecting the fault and checking the message names the line — including the issue's own verification case:

```
README.md:461: collapsed parts sum to 35 but the total says 39 -- 39 collapsed (spring ×24, tomcat ×11)
README.md:81:  block opened as #c73cf755 (line 62) but closes as #deadbeef
README.md:80:  block #c73cf755 has a story but no "env:" line — truncated mid-report
README.md:76:  stack shows 5 of 2 frames — cannot distil to more than it had
```

`ci.yml` parses as YAML with the new step in place. The script was run with GNU awk locally; the byte-safety note above is why it should also hold under `mawk` on the runner, though CI is the first place that gets exercised.

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,9 @@ jobs:
       - name: Plugin manifest version guard
         if: matrix.java == 21
         run: bash scripts/check-plugin-versions.sh
+      - name: Docs example report guard
+        if: matrix.java == 21
+        run: bash scripts/check-readme-reports.sh
       - name: Coverage summary
         if: matrix.java == 21
         run: |
--- a/scripts/check-readme-reports.sh
+++ b/scripts/check-readme-reports.sh
@@ -1,0 +1,152 @@
+#!/usr/bin/env bash
+# Guard the example reports in the docs against shapes the renderer cannot produce.
+#
+# Three separate doc bugs turned out to be one bug: prose claiming output no
+# StackDistiller run could emit, with nothing checking it. Full equality against
+# stacktale-core/src/test/resources/golden/ is too strict — the docs use a
+# different app name and shorter stacks on purpose — so this checks the
+# properties that were actually wrong (FORMAT.md §3 is the reference):
+#
+#   1. a block opening "━━━ ERROR #<id>" closes with "━━━ END #<id>", same id
+#   2. every "… N collapsed (a ×x, b ×y)" satisfies x + y + … == N
+#   3. "story (…, last N events, …)" and "stack (distilled, N of M frames)"
+#      keep their real shape, and N <= M for the stack
+#   4. a block with a story also has an "env:" line — i.e. it is not truncated
+#      mid-report while being introduced as a whole one
+#
+# On failure, fix the report in the named file. Do not adjust this script to
+# match the docs: the whole point is that the docs match the renderer.
+set -euo pipefail
+
+cd "$(dirname "$0")/.."
+
+files=(README.md)
+[ -f docs/site/index.html ] && files+=(docs/site/index.html)
+
+fail=0
+for file in "${files[@]}"; do
+  awk -v FILENAME_="$file" '
+    # Last run of digits in s. Each part reads "<framework> ×<count>", and the
+    # count is always last, so this gets it without matching the multiplication
+    # sign — which is multibyte, and mawk (the default awk on the CI runner) is
+    # byte-oriented. Also survives a framework whose name contains a digit,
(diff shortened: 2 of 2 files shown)


GabrielBBaldez (approved):
Ran all four checks against deliberately broken docs rather than trust that they fire. Each one caught its case and named the line:

```
README.md:52: collapsed parts sum to 30 but the total says 34 -- 34 collapsed (spring ×20, tomcat ×10)
README.md:81: block opened as #c73cf755 (line 62) but closes as #deadbeef
README.md:76: stack shows 99 of 2 frames — cannot distil to more than it had
README.md:81: block #c73cf755 has a story but no "env:" line — truncated mid-report
```

The first is the exact bug in the issue. The arithmetic check is the one that mattered — `39 collapsed (spring ×24, tomcat ×11)` shipped in the README for months, and 24 + 11 = 35 is the kind of thing no reviewer catches by reading.

Two decisions in here are better than what the issue asked for.

`N <= M` on the stack header was not on the list. It is the same class of impossible claim as the collapsed sum, and cheaper to check than to notice.

And the header comment says the right thing out loud:

> Do not adjust this script to match the docs: the whole point is that the docs match the renderer.

A guard without that line gets "fixed" by relaxing it the first time someone is in a hurry.

`docs/site/index.html` is included when present rather than assumed — the landing page had its own impossible version of the same line, so covering it is not incidental.

Thanks @dchaudhari7177 - the `N <= M` check and that comment are both yours, not the issue's.

PR #178, opened 2026-08-11
infra: make the build reproducible

### LINKED ISSUE #131: infra: make the build reproducible (project.build.outputTimestamp)
The build is not reproducible: `pom.xml` has no `<project.build.outputTimestamp>`, and there is no `reproducib*` anywhere in the repo.

Reproducible builds are increasingly checked for JVM libraries — it lets anyone verify that the jar on Maven Central was built from the tag it claims, which is the cheapest supply-chain assurance a project can offer. It pairs with #40 (SBOM + provenance).

## What to do

1. Add to the parent `pom.xml` properties:
   ```xml
   <project.build.outputTimestamp>2026-07-21T00:00:00Z</project.build.outputTimestamp>
   ```
   (any fixed ISO-8601 instant; convention is the release date. Maven wires it into the jar/plugin timestamps automatically.)
2. Verify with the reproducible-build plugin:
   ```bash
   mvn clean verify artifact:compare
   ```
3. If anything still differs, the usual culprits are `maven-jar-plugin` manifest entries carrying a build time and the shade plugin — both are configurable.
4. Note it in `docs/` or the README's release section, so it is discoverable by someone auditing the project.

## Verify

Two clean builds from the same commit produce byte-identical artifacts (`artifact:compare` passes).

dchaudhari7177:
Closes #131

## The problem

Two clean builds of the same commit produced jars with different checksums: every archive entry carried the wall-clock time of the build. That makes it impossible for a consumer to rebuild a tag and check the result against the jar on Maven Central — the cheapest supply-chain assurance the project can offer, and it pairs with #40 (SBOM + provenance).

## What changed

**1. One property in the parent pom.**

```xml
<project.build.outputTimestamp>2026-08-11T00:00:00Z</project.build.outputTimestamp>
```

Maven feeds this to the jar, source, javadoc and shade plugins itself, so no per-plugin configuration was needed. I checked the two usual culprits the issue mentions:

- `maven-jar-plugin` (3.5.1) adds only `Automatic-Module-Name` to the manifest — no build time.
- both shade executions (`stacktale-agent`, `stacktale-mcp`) are on 3.6.2, which honours the property.

The comment on the property says to bump it to the release date when cutting a release.

**2. A CI job that proves it, rather than trusting it to stay true.**

`Reproducible build` packages twice and diffs the jar checksums. I used that instead of `artifact:compare` (which the issue suggests) because `artifact:compare` needs a published reference release to compare against, and a `1.3.0-SNAPSHOT` on a PR branch doesn't have one. Building twice needs nothing external. The job fails loudly if no jars are found, so the diff can't pass vacuously.

**3. A `Reproducible builds` section in CONTRIBUTING.md** — what the property does, the reminder to bump it at release time, and the local two-build recipe.

## Verification — please read

**Maven is not installed in this environment, so I could not run the two-build comparison locally.** The reasoning above is from reading the poms and the plugin versions; the CI job on this PR is what actually proves the claim. If it goes red, the diff output will name the offending jar and I'll chase it.

## Note on process

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,3 +41,40 @@ jobs:
         run: |
           awk -F, 'FNR>1 {m+=$8; c+=$9} END {printf "**Line coverage: %.1f%%** (%d/%d lines)\n", c*100/(c+m), c, c+m}' \
             $(find . -path '*/site/jacoco/jacoco.csv') >> "$GITHUB_STEP_SUMMARY"
+
+  reproducible:
+    name: Reproducible build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+      - uses: actions/setup-java@v5
+        with:
+          distribution: temurin
+          java-version: 21
+          cache: maven
+      # Two clean builds of one commit must produce byte-identical jars, so a consumer can
+      # rebuild a tag and check it against Maven Central. `artifact:compare` is the other way
+      # to check this, but it needs a published reference release to compare against, which a
+      # SNAPSHOT on a PR branch does not have. Building twice needs nothing external.
+      #
+      # -DskipTests on purpose: this job checks the packaging step, and `build` already runs
+      # the suite on both JDKs.
+      - name: Build twice and compare jar checksums
+        run: |
+          sums() { find . -path '*/target/*.jar' | sort | xargs -r sha256sum; }
+          mvn -B -DskipTests clean package
+          sums > "$RUNNER_TEMP/first.txt"
+          # An empty list would make the diff below pass without checking anything.
+          if [ ! -s "$RUNNER_TEMP/first.txt" ]; then
+            echo "::error::no jars found after packaging - the check would be vacuous"
+            exit 1
+          fi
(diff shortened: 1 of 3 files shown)


GabrielBBaldez (approved):
Verified the property actually does what it claims rather than trusting that it was set — two clean `mvn package` runs of this commit:

```
build 1: f0ec4338dfc3fc92…
build 2: f0ec4338dfc3fc92…   identical
```

Building twice instead of `artifact:compare` is the right call and the comment explains why: `compare` needs a published reference release, which a SNAPSHOT on a PR branch does not have. This needs nothing external, so it works on a fork's first contribution.

`-DskipTests` on that job is also right — `build` already runs the suite on both JDKs, and this one is checking packaging.

One small thing for later, not for this PR: `prepare-release` does not move `project.build.outputTimestamp`, so every release will carry 2026-08-11. That is valid for reproducibility — a constant is exactly as reproducible as a per-release date — it just means the timestamp stops meaning "when this was built". Worth a line in the release workflow eventually; nothing to change here.

Thanks @dchaudhari7177.

PR #179, opened 2026-08-11
infra: add a link checker so README links can't rot again

### LINKED ISSUE #132: infra: add a link checker to CI so README links can't rot again

dchaudhari7177:
Closes #132

## What changed

A new `.github/workflows/links.yml`, split into two jobs — because relative and external links fail for completely unrelated reasons and shouldn't share a gate:

| Job | Trigger | Scope | Gates PRs? |
|---|---|---|---|
| `Relative links` | push + pull_request | `lychee --offline` | **yes** |
| `External links` | weekly cron + workflow_dispatch | every link | no |

The offline job is the one that would have caught the bug in the issue: the README pointing at `stacktale/src/main/java/.../StacktaleExecutors.java` after the file moved to `stacktale-core/`. It needs no network, can't flake, and fails only when a commit in this repo actually broke a path. That makes it safe to block a merge on.

The external job is scheduled rather than on-PR, for the reason the issue calls out: a 5xx from javadoc.io is not a reason to turn someone's unrelated PR red. It carries `GITHUB_TOKEN` so the ~40 github.com links in the README resolve at the API rate limit rather than the anonymous one.

`.lycheeignore` covers hosts that are fine in a browser but answer checkers badly — javadoc.io rate-limits while building a javadoc on first request, shields.io 502s under load, and the `/labels/` URLs are rendered views with no link signal in them. **The PR gate is `--offline` and never consults it**, so nothing real can hide behind it.

CONTRIBUTING.md gets a short paragraph on both jobs and the local repro command.

## Verification

I checked that the current tree passes the offline gate, so this lands green rather than importing a pre-existing failure: **all 42 relative link targets across the 10 markdown files resolve.**

**lychee itself is not installed in this environment**, so I verified that by resolving every `](path)` target against the working tree rather than by running the tool. The job on this PR is what exercises lychee — including `docs/site/index.html`, which the issue asked to include.

## Deliberately left out

A `create-issue-from-file` step so a failing scheduled run opens an issue. GitHub already emails on scheduled-workflow failure, and it would have added a third-party action for a marginal gain. Easy to add if you'd like it.

## Note on process

---

### DIFF
--- a/.github/workflows/links.yml
+++ b/.github/workflows/links.yml
@@ -1,0 +1,64 @@
+name: Links
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+  schedule:
+    # Weekly, Monday 06:17 UTC. External links rot on their own timetable, so they get
+    # their own run rather than failing an unrelated PR when a third party has an outage.
+    - cron: '17 6 * * 1'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  # Relative links only (--offline). This is the check that would have caught the README
+  # pointing at stacktale/src/main/java/.../StacktaleExecutors.java after the file moved to
+  # stacktale-core/ in the multi-module split: it is deterministic, needs no network, and
+  # fails only for something a commit in this repo actually broke. Safe to gate PRs on.
+  internal:
+    name: Relative links
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+      - name: Check relative links
+        uses: lycheeverse/lychee-action@v2
+        with:
+          args: >-
+            --offline
+            --no-progress
+            --include-verbatim
+            README.md CONTRIBUTING.md SECURITY.md CODE_OF_CONDUCT.md CHANGELOG.md
+            'docs/**/*.md'
+            docs/site/index.html
+          fail: true
+          jobSummary: true
+
+  # Everything, including external hosts. Scheduled and manual only - a 5xx from
+  # javadoc.io is not a reason to block someone's pull request.
+  external:
+    name: External links
+    if: github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+      - name: Check all links
+        uses: lycheeverse/lychee-action@v2
+        with:
+          args: >-
+            --no-progress
+            --exclude-mail
+            --include-verbatim
+            --max-retries 3
+            --retry-wait-time 5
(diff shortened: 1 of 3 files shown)


GabrielBBaldez (approved):
Verified the half you couldn't: lychee isn't installed here either, so I put the exact regression from #132 on a throwaway branch — README's `StacktaleExecutors.java` link pointed back at `stacktale/` — and let the job run against it.

```
| 🔍 Total       | 96    |
| ✅ Successful  | 29    |
| 🚫 Errors      | 1     |

### Errors in README.md
* [ERROR] <file:///.../stacktale/src/main/java/.../StacktaleExecutors.java> (at 662:28) | File not found.
```

Exit code 2, job red, pointing at the line and column. Branch deleted. That is the gate earning the right to block a merge, and it's the thing a config review alone can't tell you.

**One gap, not a blocker.** The file list covers 10 of the repo's 19 markdown files. The 9 outside it hold 3 relative links between them, and all 3 resolve today:

- `.github/actions/report/README.md` → `../../../docs/FORMAT.md`, `../../../README.md`
- `plugins/stacktale/README.md` → `../../docs/mcp-setup.md`

Those `../../../` paths are the fragile kind — they break silently when a file moves, which is the failure this issue was about. Worth widening later; I'll open a follow-up rather than hold this.

Skipping `create-issue-from-file` was the right call for the reason you gave.

On your process note — the issue was claimed, and that's mine to sort out, not yours. Handling it on #132; it doesn't reflect on the work here.

Thanks for splitting this into two jobs instead of pointing one checker at everything. A deterministic gate plus a flaky check on a cron is the difference between a link checker that survives and one that gets disabled after the third unrelated red build.

Issue #126, opened 2026-07-25
test: ReportPipeline has no direct unit test — the rollback/confirm state machine is untested


GabrielBBaldez:
`ReportPipeline` owns the never-throw guarantee and the dedup/storm state machine, and it has **no direct unit test**. Per the checked-in JaCoCo report: `ReportPipeline` 0% (110 lines missed), `ReportPipeline.Settings.Builder` 0% (42), `UncaughtHandler` 0%, `AgentCaptures` 0%, `StacktaleExecutors` 0%. There is no `ReportPipelineTest` in `stacktale-core/src/test/`.

Everything is reached only end-to-end through the Logback tests.

## Why this is the coverage that matters

The rollback/confirm protocol from #51/#57 (`process():239-282`) is a four-state machine: storm `SUPPRESS` → `rollback`; `STORM_LINE` → `confirmStormLine` + `rollback`; write failure → `rollback` + rethrow; success → `confirmReport`.

Its failure mode is silent. A regression that skips a `rollback` leaves `reportPending=true`, and `Deduper.decide():66-68` then returns `SILENT` for that fingerprint **until the 5-minute window rolls over** — an error that simply stops being reported, with no test that would go red.

One such gap is reachable today: a throwing `writer.append` in the `STORM_LINE` branch at `:245` skips both `confirmStormLine` and `rollback`, producing exactly that ~5-minute blind spot.

## What to do

Add `ReportPipelineTest` in `stacktale-core` driving a fake `Host` and an injectable failing writer:

- one report → rollback → a fresh report is still emitted
- storm suppress → rollback
- storm-line write failure → the next occurrence still reports (this one should fail before the fix)
- `close()` drains pending counters

Separately, `LogbackXmlConfigTest.configuresEveryPropertyFromXmlAndReports` omits six knobs — `echoSuppressionMillis`, `containerLogger`, `emitReportsToLogger`, `maxReportsPerMinute`, `format`, `redactionCorrelation` — so a Joran setter-name typo in those would not be caught, despite the CONTRIBUTING rule that every new property gets XML coverage. Add them.

## Verify

`mvn -pl stacktale-core verify`, and the mutation score on `ReportPipeline` before/after (`mvn -pl stacktale-core test-compile org.pitest:pitest-maven:mutationCoverage`).

GabrielBBaldez:
Both halves of this landed and nothing here is outstanding.

[...]

Issue #57, opened 2026-07-13
Write-durability & concurrency hardening (dedup/storm/rotation edges)


GabrielBBaldez:
Found in deep review — a cluster of small, verified robustness fixes, worth doing before 1.0 (each cheap):
- \`Deduper.rollback\` does \`stats.remove()\`, wiping \`firstSeen\`/\`total\` → resets the \"seen N× this session\" recurrence after a transient write failure. Reset only window fields instead.
- StormLimiter zeroes suppressedSinceLine **before** the storm line is durably written → count lost on write failure. Mirror Deduper.confirmWritten.
- A \`repeated N×\` line can be appended **before** its own report block (decide→append not atomic) → document id-binding in FORMAT.md and make StReportFile tolerate it.
- Two JVMs sharing errors-ai.log: concurrent rotation can delete the backup / truncate. Document the single-writer assumption and replace \`size==0 → TRUNCATE_EXISTING\` with non-destructive create.
- Windows: rotation fails while a reader holds the file → reporting silently disabled until the handle closes. Degrade (append past cap + retry, or copy-truncate) and rate-limit the warning.
- Torn block / half NDJSON line on disk-full: document in FORMAT.md §7 that parsers must skip unparseable lines (§3 already says this for text).

GabrielBBaldez:
Done (`cbdf39f`): the four low-risk items —
- [x] `Deduper.rollback` keeps `firstSeen`/`total`
- [x] `StormLimiter` clears the suppressed count only after a durable write (`confirmStormLine`)
- [x] FORMAT.md: `repeated Nx` id-binding + parsers tolerate an unseen report (StReportFile already does)
- [x] FORMAT.md §7: parsers skip unparseable NDJSON lines

Remaining, for a focused pass:
- [ ] concurrent cross-JVM rotation (single-writer doc + non-destructive create)
- [ ] Windows: rotation while a reader holds the file (degrade + rate-limit the warning)

GabrielBBaldez:
Done (`59068a0`) — the last two rotation items, closing this out:

- [x] **Concurrent cross-JVM rotation.** `size==0` no longer writes the header with `TRUNCATE_EXISTING`; it appends it. No write path truncates anymore, so a second writer (or a race) can interleave or repeat the header but can't wipe another process's data. The single-writer assumption is documented on `ReportWriter` (and already in `SECURITY.md`).
[...]

Issue #51, opened 2026-07-13
Storm-suppressed report leaves dedup orphaned: "repeated N×" lines with no report


GabrielBBaldez:
Found in deep review (verified). \`Deduper.decide()\` commits REPORT state (count=1, lastReport=now — Deduper.java:51-56) **before** the storm limiter is consulted. When the limiter returns SUPPRESS/STORM_LINE, \`process()\` returns without \`deduper.rollback()\` (ReportPipeline.java:229-232), unlike the write-failure path which does. Result: a new error first seen during a storm never gets a full report; later occurrences within the dedup window take the SUMMARY path (which bypasses the storm gate) and write \`━ #id repeated N× ━\` lines referencing an id that exists nowhere — contradicting FORMAT.md §5. The MCP parser (StReportFile) then drops the error entirely.

Fix: track suppression in Deduper (\`pendingReport\` flag) so the next decide() returns REPORT again (re-gated) instead of SUMMARY, cleared only after a durable write. Add a pipeline test.

PR #180, opened 2026-08-11
fix(core): roll back the dedup decision when a storm line fails to write

dchaudhari7177:
Refs #126

## The bug

The issue names a gap that's reachable today, and it is real:

```java
if (storm.action() == StormLimiter.Action.STORM_LINE) {
    writer.append(renderer.stormLine(...));   // throws
    stormLimiter.confirmStormLine(...);       // skipped
    deduper.rollback(fingerprint);            // skipped
    return;
}
```

With the rollback sitting *after* the append, a failing storm-line write skips it and leaves `reportPending = true` for that fingerprint. `Deduper.decide` then answers `SILENT` for it (`Deduper.java:66-68`) until the dedup window rolls over — **the error stops being reported for ~5 minutes on the default window, and nothing goes red.**

## The fix

Move the rollback into a `finally`:

```java
try {
    writer.append(renderer.stormLine(...));
    stormLimiter.confirmStormLine(storm.suppressed()); // #57: clear only once written
} finally {
    deduper.rollback(fingerprint);
}
```

That is the invariant #51 actually wanted. The rollback is about *this error's own* report, which was never written on this path whether or not the storm line itself made it — so it should be unconditional. `confirmStormLine` stays inside the `try`, so #57's "clear only once written" still holds. The throw still reaches the outer handler, so the never-throw guarantee and the consecutive-failure parking are unchanged.

## The tests

New `ReportPipelineTest` in `stacktale-core`, asserting on **what reached the file** — the only thing a user of stacktale can observe:

- a written report is confirmed, and the repeat becomes a summary
- a storm-suppressed report is rolled back and still reports after the window
- **a failed storm-line write still lets the next occurrence report** ← the gap
- a failed storm line keeps its suppressed count for the next line (#57)
- `close()` drains the pending summary counts
- `process()` never throws, and parks after repeated write failures

**Red-before-green, per the TDD rule in CONTRIBUTING:** with the rollback moved back after the append, `aFailedStormLineWriteStillLetsTheNextOccurrenceReport` fails (1 report on file, not 2) and the other five still pass.

One detail worth flagging for future test authors: assertions are anchored to line starts, because the file header *documents its own delimiters* — a plain `contains("━━━ ERROR #")` matches the header comment as well as a real report. I got that wrong first.

## Two seams, both package-private

- **`forTesting(...)`** — `create()` cannot build a pipeline that is both active and unable to write: `ReportWriter`'s constructor probes the destination, and a failed probe becomes `isActive() == false`. Injecting the collaborators is the only way to reach the failure branches.
- **an injectable clock**, threaded through to `Deduper` and `StormLimiter`. Their windows are minutes long. `DeduperTest` and `StormLimiterTest` already inject a clock, so this follows the existing convention rather than inventing one.

The failure is injected through a `Renderer` that throws on `stormLine`, not a subclassed `ReportWriter` — `ReportWriter` is `final`, and the pipeline can't tell the two apart: both leave `STORM_LINE` decided with nothing on disk. Say the word if you'd rather `ReportWriter` grew a seam instead.

## Verification — please read

**Maven is not installed in this environment.** I compiled and ran the suite against the JDK directly (JUnit 6.1.1 + AssertJ 3.27.7 from the local repository, against the pinned 6.1.3).

- **117 of the 122 runnable core tests pass.**
- 5 fail: `EnvCollectorTest` ×1, `ReportRendererTest` ×4. **These fail identically on an unmodified `main` in the same setup** — I built pristine `main` from `git archive` to confirm it. They need Maven's resource filtering and the pinned 6.1.3.
- `JsonReportRendererTest` couldn't be compiled at all without Jackson.

So: the new tests and every test that could run are green, and the 5 failures are artifacts of my environment, not of this change. **CI is the real check.** I also could not run PIT for the before/after mutation score the issue asks for.

## Scope

Deliberately limited to `stacktale-core`. The issue's second half — the six knobs `LogbackXmlConfigTest` omits (`echoSuppressionMillis`, `containerLogger`, `emitReportsToLogger`, `maxReportsPerMinute`, `format`, `redactionCorrelation`) — lives in the `stacktale` module and is a separate logical change per the one-change-per-PR rule. Logback isn't available here to run it against either. Marked `Refs` rather than `Closes` for that reason; happy to follow up.

## Note on process

---

### DIFF
--- a/stacktale-core/src/main/java/io/github/gabrielbbaldez/stacktale/ReportPipeline.java
+++ b/stacktale-core/src/main/java/io/github/gabrielbbaldez/stacktale/ReportPipeline.java
@@ -6,6 +6,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.LongSupplier;
 import java.util.regex.Pattern;
 
 /**
@@ -218,6 +219,11 @@ public final class ReportPipeline {
             };
 
     private ReportPipeline(Settings settings, Host host, ReportWriter writer, Renderer renderer) {
+        this(settings, host, writer, renderer, System::currentTimeMillis);
+    }
+
+    private ReportPipeline(Settings settings, Host host, ReportWriter writer, Renderer renderer,
+                           LongSupplier clock) {
         this.settings = settings;
         this.host = host;
         this.renderer = renderer;
@@ -225,9 +231,9 @@ public final class ReportPipeline {
         this.storyBuffer = new StoryBuffer(settings.storySize(), settings.storyWindowMillis(),
                 settings.correlationMdcKeys(), 200);
         this.distiller = new StackDistiller(settings.appPackages());
-        this.deduper = new Deduper(settings.dedupWindowMillis(), 60_000, System::currentTimeMillis);
+        this.deduper = new Deduper(settings.dedupWindowMillis(), 60_000, clock);
         this.stormLimiter = settings.maxReportsPerMinute() > 0
-                ? new StormLimiter(settings.maxReportsPerMinute(), 60_000, 10_000, System::currentTimeMillis)
+                ? new StormLimiter(settings.maxReportsPerMinute(), 60_000, 10_000, clock)
                 : StormLimiter.disabled();
         this.env = new EnvCollector(Thread.currentThread().getContextClassLoader(),
                 settings.appName(),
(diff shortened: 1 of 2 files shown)


GabrielBBaldez (approved):
Real, and proven rather than argued. I undid only the `finally` — leaving the rest of the PR in place — and ran your own tests against that:

```
aFailedStormLineWriteStillLetsTheNextOccurrenceReport   expected: 2  but was: 1
```

The second occurrence never reports. That is the ~5 minutes of silence in the description, and no other test in the suite noticed it.

The `finally` is the right shape for the reason the comment gives: this error's own report was never written, and that is true whether or not the storm line made it. Rollback is not conditional on the thing that failed.

Two decisions worth calling out.

**The clock seam.** Threading a `LongSupplier` through a private constructor rather than reaching for a mock is what makes the window-rollover branches testable at all — `Deduper` and `StormLimiter` already took one, so the pipeline was the only thing standing between the tests and them.

**`confirmStormLine` stays inside the `try`, not the `finally`.** Those two look symmetric and are not: the suppressed count must survive a failed write so the next line reports it, while the dedup decision must be unwound. `aFailedStormLineKeepsItsSuppressedCountForTheNextLine` pins that, and getting it backwards would have been easy and silent.

`mvn verify` green across all 8 modules here.

Thanks @dchaudhari7177 — #126 asked for tests and you found a bug writing them, which is the outcome that issue was hoping for.

PR #184, opened 2026-08-13
ci: check links in all markdown files

### LINKED ISSUE #181: infra: the link checker covers 10 of the repo's 19 markdown files
#179 added the link checker, and it works — I broke a README path deliberately and the job went red on the right line. But it only looks at 10 of the repo's 19 markdown files.

The file list in `.github/workflows/links.yml` is spelled out explicitly, twice (once per job):

```
README.md CONTRIBUTING.md SECURITY.md CODE_OF_CONDUCT.md CHANGELOG.md
'docs/**/*.md'
docs/site/index.html
```

Nothing under `examples/`, `plugins/` or `.github/` is covered. Between them those hold three relative links, and all three resolve today, so nothing is broken right now:

- `.github/actions/report/README.md` → `../../../docs/FORMAT.md` and `../../../README.md`
- `plugins/stacktale/README.md` → `../../docs/mcp-setup.md`

Those `../../../` paths are exactly the fragile kind. They're the same failure #132 was about — a file moves, the link rots silently, and nobody notices until a reader clicks it.

## What to change

Replace the explicit list with a repo-wide glob in **both** jobs, minus the directories that shouldn't be walked:

```
'**/*.md' docs/site/index.html
--exclude-path target
--exclude-path node_modules
```

`target/` matters: a local `mvn package` leaves generated markdown in module `target/` directories, and without excluding it the job's result depends on whether the runner happens to have built first. `node_modules` is the same idea for anyone who runs the site tooling locally.

Two things, and the second is the one people skip:

1. **It still passes on a clean tree.** The whole point is landing green rather than importing a pre-existing failure.
2. **It still fails when a link actually breaks — in one of the newly covered files.** Point `plugins/stacktale/README.md` at `../../docs/mcp-setup-typo.md` on a scratch branch, push, and confirm the job goes red naming that file. A wider glob that silently matches nothing would still be green on step 1, which is why step 1 alone proves nothing.

Say the word here and it's yours.

dongyikuan919:
## Summary
- expand the relative and external link checks to every Markdown file in the repository
- exclude generated `target` output and local `node_modules` content from both jobs

Closes #181

## Validation
- YAML parsed successfully with Ruby
- verified the repository currently contains 19 Markdown files covered by the glob
- `git diff --check`

`lychee` is not installed locally; the workflow itself will run the authoritative check.

### DIFF
--- a/.github/workflows/links.yml
+++ b/.github/workflows/links.yml
@@ -30,9 +30,10 @@ jobs:
             --offline
             --no-progress
             --include-verbatim
-            README.md CONTRIBUTING.md SECURITY.md CODE_OF_CONDUCT.md CHANGELOG.md
-            'docs/**/*.md'
+            '**/*.md'
             docs/site/index.html
+            --exclude-path target
+            --exclude-path node_modules
           fail: true
           jobSummary: true
 
@@ -53,9 +54,10 @@ jobs:
             --include-verbatim
             --max-retries 3
             --retry-wait-time 5
-            README.md CONTRIBUTING.md SECURITY.md CODE_OF_CONDUCT.md CHANGELOG.md
-            'docs/**/*.md'
+            '**/*.md'
             docs/site/index.html
+            --exclude-path target
+            --exclude-path node_modules
           fail: true
           jobSummary: true
         env:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,8 +45,8 @@ that check fails after a dependency bump, update the table in `README.md`.
 
 Markdown links are checked too, by `.github/workflows/links.yml`:
 
-- **On every PR**, `lychee --offline` verifies the *relative* links in the
-  top-level docs and `docs/`. This is the one that blocks a merge — it is
+- **On every PR**, `lychee --offline` verifies the *relative* links in every
+  Markdown file in the repository. This is the one that blocks a merge — it is
   deterministic and fails only when a commit here actually broke a path (e.g.
   moving a file that the README points at).
 - **Weekly**, a separate job checks external links as well. Third-party outages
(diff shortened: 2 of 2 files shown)


GabrielBBaldez (approved):
Ran the second check from the issue, since it's the one that separates a wider glob from a glob that quietly matches nothing — both are green on a clean tree.

First the counts moved, which a no-op glob wouldn't do:

| | Total | Checked (relative) |
|---|---|---|
| `main` | 96 | 30 |
| this PR | 102 | 31 |

Then the direct proof. I pointed `plugins/stacktale/README.md` — one of the nine files that had no coverage before — at a path that doesn't exist, on a throwaway branch:

```
| 🔍 Total       | 102   |
| ✅ Successful  | 30    |
| 🚫 Errors      | 1     |

### Errors in plugins/stacktale/README.md
* [ERROR] <file:///.../docs/mcp-setup-NAO-EXISTE.md> (at 44:11) | File not found.
```

Exit code 2, and it names the newly covered file. Branch deleted.

That also explains why only one relative link was added rather than the three the issue counted. Breaking that one link moved *Successful* 31 → 30, so it was among the checked set; the other two (`../../../README.md` and `../../../docs/FORMAT.md`) resolve to files already in the run, and lychee counts a resolved target once. Nothing missing — the arithmetic just isn't one-link-per-occurrence.

Excluding `target` matters more than it looks, and I'd have accepted this PR without it: after a local `mvn package` the tree carries generated markdown, so leaving it in would make the result depend on whether the machine had built first. Green on CI, red on a contributor's laptop, for reasons neither could see.

One thing was left behind, and I pushed it to your branch rather than bounce the PR for six lines: `CONTRIBUTING.md` still described the check as covering "the top-level docs and `docs/`" and still listed the old explicit file set as the local repro command — including a line telling the next person to keep that list in step with this job. Now updated to match.

PR #187, opened 2026-08-19
feat: add Quarkus extension

### LINKED ISSUE #82: stacktale-quarkus: Quarkus extension for zero-config reports
Quarkus apps can already use stacktale **today**: Quarkus logs through the **JBoss LogManager**, which is a `java.util.logging` implementation — so the `stacktale-jul` handler attaches to it (or, with the Logback extension, the `stacktale` appender). Manual config, but it works.

This tracks the idiomatic layer. Unlike Spring/Micronaut, the Quarkus-native shape is a **Quarkus extension** (build-time processing), which is also GraalVM-native-friendly — and stacktale already ships native metadata (#45):

- **Extension structure** — `runtime` + `deployment` modules; a build step that registers the stacktale JUL handler (or Logback appender) at startup with zero user config.
- **Config** — `stacktale.*` via `@ConfigMapping` / `application.properties`, parity with the other starters.
- **HTTP request-line into the story** — a Quarkus request filter (`@ServerRequestFilter` / Vert.x) opening each story with the request line.
- **Native** — verify under `quarkus build --native`, reusing the RuntimeHints / `resource-config.json` from #45.
- appPackages deduction; sample app + IT; docs.

Reuses `stacktale-core` + `stacktale-jul` — this module is the Quarkus glue. Likely a new repo `stacktale/stacktale-quarkus`.

DenizAltunkapan:
## What & why

Adds `stacktale-quarkus`, a Quarkus extension that gives Quarkus apps zero-config stacktale reports: add the dependency and every logged error becomes an st/1 report in errors-ai.log, with no logging.properties or logback.xml to touch. Quarkus logs through the JBoss LogManager (a java.util.logging implementation), so the extension reuses the existing stacktale-jul handler and adds only the Quarkus glue. It uses a runtime plus deployment split so the wiring is decided at build time and stays native-friendly.

Closes #82

## How it was verified

Verified through the Quarkus reactor. The deployment module's StacktaleExtensionTest boots the extension in-process with QuarkusUnitTest and asserts a real st/1 report is written, with the root cause unwrapped and the app frame marked. No separate example app needed. Built against Quarkus 3.15.1 and stacktale-core 1.3.0-SNAPSHOT.

## Checklist

- [x] The issue was claimed with a comment before starting. I did not claim it beforehand, opening the PR anyway as CONTRIBUTING suggests.
- [x] One logical change (a single new module, nothing else touched)
- [x] Behavior changes arrive with the test that demanded them (StacktaleExtensionTest)
- [x] No golden-file test changed. The st/1 output comes from the existing core, unchanged.
- [x] New config property? The stacktale.* keys are @ConfigMapping methods with defaults that preserve current behavior, exercised by StacktaleExtensionTest and documented in the module README. LogbackXmlConfigTest and the README config table apply to the Logback appender, not this JUL-based extension.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,14 +30,21 @@ that way deliberately.
   arguments — `long` stays `long` instead of becoming `java.lang.Long`. Where the method
   can't be resolved unambiguously (overloads with the same arity), it falls back to the
   runtime class, which still names something a call can be written against. Only the
-  innermost frame becomes the seed:
-  a seed naming the outer caller would describe a call that did not fail. **Off by
-  default**, and worth leaving off for most projects — it is the only section that renders
+  innermost frame becomes the seed: a seed naming the outer caller would describe a call
+  that did not fail. **Off by default**, and worth leaving off for most projects — it is the only section that renders
   argument values against a named signature, which is a larger privacy surface than the
   rest of a report combined. Values go through the same redaction as everything else, and
   the name and value are cleaned as one string so a secret-named parameter masks its value.
   An older agent paired with a newer core loses the seed and keeps `captured:`, rather than
   failing. Additive under the FORMAT §6 rules, so `st/1` does not change version. (#135)
+- **`stacktale-quarkus`** — a Quarkus extension, so a Quarkus app gets reports by adding the
+  dependency, with no `logging.properties` to edit. Quarkus logs through the JBoss LogManager,
+  which is a `java.util.logging` implementation, so the extension reuses the `stacktale-jul`
+  handler and adds only the Quarkus wiring: a runtime/deployment split that decides the
+  configuration at build time and stays native-image friendly, `stacktale.*` keys through
+  `@ConfigMapping`, and a `@ServerRequestFilter` that opens each request's story with its HTTP
+  line — the counterpart of the Spring starter's servlet and WebFlux filters. Thanks
+  @DenizAltunkapan. (#82)
 - **Reproducible builds.** Two clean builds of the same commit now produce byte-identical
   jars, so a tag can be rebuilt and checked against what is on Maven Central. CI packages
   twice and diffs the checksums on every PR. (#178)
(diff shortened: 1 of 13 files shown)


GabrielBBaldez (approved):
Checked the thing a new module most easily gets wrong — being in the tree but not in the build — and it holds. `stacktale-quarkus` is in the reactor's `<modules>`, both submodules build in CI, and the test genuinely boots Quarkus rather than asserting against a mock:

```
INFO  [io.quarkus] stacktale-quarkus-deployment on JVM (powered by Quarkus 3.15.1) started
INFO  [io.quarkus] Installed features: [cdi, stacktale]
SEVERE [StacktaleExtensionTest] Failed to confirm order 999: IllegalStateException: customer cache returned null
INFO  [stacktale] AI error report #15111260 → .../errors-ai-test.log
```

`Installed features: [cdi, stacktale]` is the line that matters — the extension is really registered by the build step, not merely on the classpath — and a report reaches the file at the end of it.

Reusing the JUL handler rather than writing a fourth one is the right call and worth saying out loud: JBoss LogManager *is* JUL, so a separate handler would have been a second copy of logic that already has the most configuration surface in the project. Both poms declare `stacktale.module.name`, so the jars keep stable JPMS names like every other module — easy to forget in a submodule that inherits from a new parent.

I did look hard at `StacktaleRequestFilter`, because it logs the raw query string and a request line ending in `?token=…` would land in the story. It is not a leak: the Spring servlet filter builds the same line the same way, and `ReportRenderer` puts every story message through `clean()`, so a secret-shaped value is redacted there like anywhere else. Consistent with the existing adapters rather than a new hole.

**Pushed to your branch** rather than bouncing the PR, since none of it is about your code:

- the README described Quarkus as roadmap in two places (`#82`, "on the roadmap", "what's open next") while this ships it;
- it was missing from the ecosystem table, so nobody reading the README would find the module;
- and from the JPMS table, which lists every other artifact;
- plus a `[Unreleased]` changelog entry — a new backend is exactly what that section exists for.

**One follow-up, not a blocker:** #188. Every other adapter has a floor leg in the compatibility matrix and this one pins 3.15.1 with nothing testing an older Quarkus — which matters more here than elsewhere, since the runtime/deployment split and the extension descriptor are build-time contracts that move between minors.

Thanks for splitting runtime and deployment properly instead of putting the wiring in a single module. That is the part people skip because it works either way on the JVM, and it is what keeps native image on the table.

PR #199, opened 2026-09-03
test(logback): cover the six XML properties the config test omitted (#126)

dchaudhari7177:
Re #126 — the second half of it.

## Scope: why only half

The first half of #126 is **already done on `main`**. `ReportPipelineTest` exists with the four cases the issue lists, and the storm-line gap it describes (`writer.append` throwing at the `STORM_LINE` branch, skipping both `confirmStormLine` and `rollback`) is fixed — the `rollback` now sits in a `finally`, with a comment naming #51. So there was nothing left to add there.

What is still open is the paragraph at the end:

> `LogbackXmlConfigTest.configuresEveryPropertyFromXmlAndReports` omits six knobs — `echoSuppressionMillis`, `containerLogger`, `emitReportsToLogger`, `maxReportsPerMinute`, `format`, `redactionCorrelation` — so a Joran setter-name typo in those would not be caught.

Confirmed: all six were absent from the XML. This PR is that.

## The assertion matters more than the six elements

Adding the elements alone would not actually have fixed the problem, and this is the part worth reviewing.

**Joran does not throw on an element it cannot bind.** It records a WARN on the context and carries on with the property unset. The existing test only reads the report file afterwards — so with a misspelled setter it still finds `gateway timeout` in the log, still passes, and the knob silently does nothing. That is precisely how six properties could sit uncovered without anyone noticing.

So `assertNoJoranComplaints(ctx)` fails on any WARN/ERROR status. That:

- covers **every** property in the XML at once, not just the six
- covers a property added **later**, the moment it appears in the XML, without anyone having to invent an observable effect for it

## Verified the guard is real

Misspelled one property as `echoSupressionMillis` (one `s`) and re-ran:

```
[ERROR] LogbackXmlConfigTest.configuresEveryPropertyFromXmlAndReports:79
  Joran could not bind part of the XML — a setter name or type is wrong:
  [WARN ... Ignoring unknown property [echoSupressionMillis]
   in [io.github.gabrielbbaldez.stacktale.logback.StacktaleAppender]]
```

Restored, and it passes again.

## Two properties get their own test

- **`format`** changes the shape of every line, so it can't share assertions with the text-format block. Its test asserts the output actually `startsWith("{")` — a name-only check would pass even if `setFormat` stopped selecting the JSON renderer.
- **`emitReportsToLogger`** routes reports back through logback, so it would show up in the sibling appenders of any context it shared.

## `anUnknownPropertyIsReportedByJoran` is the control

Without it, `assertNoJoranComplaints` could be passing simply because Joran never complains about anything in this setup — which would make every other use of it vacuous. This test asserts the mechanism fires.

## Verification

`mvn -pl stacktale -am test` → **BUILD SUCCESS**, 128 core tests + 32 module tests, 0 failures. `LogbackXmlConfigTest` itself: 4 run, 0 failures.

## Process note

CONTRIBUTING asks for an issue claim before writing code, and I didn't post one — the work was already done when I read that section, and it seemed worse to claim an issue I'd finished than to just say so here. Happy to follow the claim-first flow on anything further. If someone else has the remaining `ReportPipeline` work in progress, this half is independent of it (different module, different file) and shouldn't collide.

### DIFF
--- a/stacktale/src/test/java/io/github/gabrielbbaldez/stacktale/LogbackXmlConfigTest.java
+++ b/stacktale/src/test/java/io/github/gabrielbbaldez/stacktale/LogbackXmlConfigTest.java
@@ -2,6 +2,7 @@ package io.github.gabrielbbaldez.stacktale;
 
 import ch.qos.logback.classic.LoggerContext;
 import ch.qos.logback.classic.joran.JoranConfigurator;
+import ch.qos.logback.core.status.Status;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.TempDir;
@@ -11,6 +12,7 @@ import java.io.ByteArrayInputStream;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.List;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
@@ -47,8 +49,12 @@ class LogbackXmlConfigTest {
                     <captureExceptionFields>true</captureExceptionFields>
                     <redactionEnabled>true</redactionEnabled>
                     <redactPattern>BR\\d{2}-\\d{4}</redactPattern>
+                    <redactionCorrelation>true</redactionCorrelation>
                     <correlationMdcKeys>traceId</correlationMdcKeys>
                     <zone>UTC</zone>
+                    <echoSuppressionMillis>250</echoSuppressionMillis>
+                    <containerLogger>com.acme.container</containerLogger>
+                    <maxReportsPerMinute>50</maxReportsPerMinute>
                   </appender>
                   <root level="INFO">
                     <appender-ref ref="STACKTALE"/>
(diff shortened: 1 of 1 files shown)


GabrielBBaldez (approved):
Reproduced your experiment rather than taking it on trust — misspelled `echoSuppressionMillis` as `echoSupressionMillis` and the guard caught it, naming the property:

```
java.lang.AssertionError: Joran could not bind part of the XML — a setter name or type is wrong:
  [WARN in ch.qos.logback.core.model.processor.ImplicitModelHandler -
   Ignoring unknown property [echoSupressionMillis] in [...logback.StacktaleAppender]]
```

Also checked the premise: all six really were absent from `main`. The one that looked present, `format`, was `.formatted(...)` on the XML string, not a `<format>` element.

**The insight is the part worth keeping.** Adding the six elements is the obvious reading of the issue and it would have left the hole exactly where it was, because Joran records a WARN and carries on rather than throwing — the report file still contains what the assertions look for. Asserting on the status list turns "did the config bind" into something the test can see, and it covers the next property for free. Scoping to the half of #126 still open, and saying so, saved me re-reading the first half.

**One thing worth knowing, not a blocker.** The assertion catches *any* status at WARN or above, and the appender itself writes there: `Host.warn` in `StacktaleAppender` calls `addWarn(...)`, which lands in the same `StatusManager` this reads. Nothing in this test triggers it, so it passes — but if a future test in this class ever exercises a path where the appender legitimately warns, the failure would read as "a setter name or type is wrong" while being nothing of the sort. Narrowing the message, or filtering to statuses whose origin is a Joran model handler, would keep that from sending someone down the wrong path. Happy for it to be a follow-up rather than churn on this PR.

Thanks for reading the issue as a claim to verify instead of a task list — checking whether the first half was already done, and then finding that the stated fix would not have worked, is the part that made this worth more than six lines of XML.

PR #230, opened 2026-09-05
fix(test): narrow assertNoJoranComplaints to Joran binding failures

### LINKED ISSUE #205: test: the Joran guard blames the XML for warnings the appender raised itself
Follow-up to #199, which added `assertNoJoranComplaints` to `LogbackXmlConfigTest`. The guard is right and it earns its place — it catches a setter-name typo that the old assertions would have missed. This is about what it says when it fires for a different reason.

It fails on any status at WARN or above:

```java
List<Status> bad = ctx.getStatusManager().getCopyOfStatusList().stream()
        .filter(s -> s.getEffectiveLevel() >= Status.WARN)
        .toList();
assertThat(bad)
        .withFailMessage("Joran could not bind part of the XML — a setter name or type is wrong: %s", bad)
```

The appender writes to that same status manager. `StacktaleAppender`'s `ReportPipeline.Host` implements `warn` as:

```java
public void warn(String message, Throwable t) {
    addWarn(message, t);
}
```

`addWarn` is `ContextAwareBase`'s, so it lands in the `LoggerContext`'s `StatusManager` — the list this assertion reads.

Nothing in the class triggers it today, which is why the test passes. But the pipeline calls `Host.warn` on real runtime problems: a write that fails, a rotation that cannot proceed. The moment a test in this class exercises one of those, the failure reads

> Joran could not bind part of the XML — a setter name or type is wrong

about a warning that has nothing to do with Joran, and sends the reader to check setter names that are perfectly fine.

## What to change

Either is fine:

- **Narrow the filter.** Joran's binding complaints come from Logback's own model handlers — `ImplicitModelHandler` and friends — so filtering on the status origin keeps the assertion about configuration and lets the appender's own warnings through.
- **Or widen the message.** Keep catching everything, and say what is actually true: that something warned during configuration, showing the status, without asserting which kind it was.

The first is more precise; the second is smaller and honest. A maintainer's preference either way is fine.

- The typo case still fails: misspell `echoSuppressionMillis` as `echoSupressionMillis` in the XML and confirm the test still catches it.
- A warning from the appender no longer claims to be a binding error. The cheapest way to produce one is a `<file>` path the writer cannot create — point it at a directory that does not exist and is not creatable — then assert on whatever the new behaviour should be.

Say the word here and it's yours.

OswaldOniSango:
StacktaleAppender's own warnings (write failures, rotation issues) were landing in the same StatusManager that assertNoJoranComplaints inspects, so a real appender warning would be reported as a Joran binding error. Exclude statuses whose origin is StacktaleAppender, since those describe appender behaviour rather than XML binding problems, and add a test covering that case.

## What & why

`assertNoJoranComplaints` treated all WARN/ERROR statuses as Joran binding failures, including diagnostics emitted by `StacktaleAppender`. This change excludes statuses originating from the appender and adds a regression test for an appender warning. Production behavior is unchanged.

Closes #205

## How it was verified

The new regression test failed against the original helper, reproducing the misleading Joran binding error.

* [x] `mvn verify` is green (JDK 17+)

## Checklist

* [x] The issue was claimed with a comment before starting.
* [x] One logical change.
* [x] Bug fixes start from a failing test.
* [x] No golden-file test changed.
* New config property: not applicable.

### DIFF
--- a/stacktale/src/test/java/io/github/gabrielbbaldez/stacktale/LogbackXmlConfigTest.java
+++ b/stacktale/src/test/java/io/github/gabrielbbaldez/stacktale/LogbackXmlConfigTest.java
@@ -3,7 +3,6 @@ package io.github.gabrielbbaldez.stacktale;
 import ch.qos.logback.classic.LoggerContext;
 import ch.qos.logback.classic.joran.JoranConfigurator;
 import ch.qos.logback.core.status.Status;
-import io.github.gabrielbbaldez.stacktale.logback.StacktaleAppender;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.TempDir;
@@ -16,20 +15,24 @@ import java.nio.file.Path;
 import java.util.List;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.assertj.core.api.AssertionsForClassTypes.assertThatThrownBy;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 /**
  * Exercises the EXACT adoption path from the README: configuring the appender through
  * logback XML (Joran calls the public setters). If a setter name or type is wrong, this
  * test fails even though programmatic tests pass.
  *
- * <p>The Joran diagnostic check excludes statuses originating from StacktaleAppender:
- * they describe appender validation or operational problems, not Joran binding failures.
- * Those problems must be checked by the tests exercising that behaviour.
+ * <p>The Joran diagnostic check only looks at statuses recorded while {@code doConfigure}
+ * was running. Everything that belongs to configuration lands there — Joran's own binding
+ * complaints, but also {@code StacktaleAppender.start()} rejecting a bad {@code <zone>} or
+ * {@code <redactPattern>} — because {@code start()} runs synchronously inside {@code
+ * doConfigure}. A write or rotation that fails once events are flowing is appended later,
+ * after {@code doConfigure} has already returned, and falls outside that prefix.
  */
 class LogbackXmlConfigTest {
 
     private LoggerContext ctx;
+    private int configEnd;
 
     @AfterEach
     void tearDown() {
@@ -71,11 +74,7 @@ class LogbackXmlConfigTest {
                 </configuration>
                 """.formatted(file.toString().replace("\\", "/"));
 
-        ctx = new LoggerContext();
-        ctx.setMDCAdapter(MDC.getMDCAdapter());
-        JoranConfigurator joran = new JoranConfigurator();
-        joran.setContext(ctx);
-        joran.doConfigure(new ByteArrayInputStream(xml.getBytes(StandardCharsets.UTF_8)));
+        configure(xml);
 
         org.slf4j.Logger log = ctx.getLogger("com.acme.CheckoutService");
         log.info("charging card for order 42");
@@ -85,7 +84,7 @@ class LogbackXmlConfigTest {
                 .withFailMessage("XML-configured appender produced no report file — Joran/setter wiring is broken. Context status: %s",
                         ctx.getStatusManager().getCopyOfStatusList())
                 .isTrue();
-        assertNoJoranComplaints(ctx);
+        assertNoJoranComplaints(ctx, configEnd);
 
         String content = Files.readString(file, StandardCharsets.UTF_8);
         assertThat(content).contains("IllegalStateException: gateway timeout");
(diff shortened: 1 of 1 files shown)


GabrielBBaldez (changes requested):
Thanks for taking this, and for extending `anUnknownPropertyIsReportedByJoran` to assert the helper actually throws — that addition is what stops the narrowed guard from quietly becoming vacuous, and it's easy to leave out.

The origin filter has a hole, though, and it's worth walking through because it has the same shape as the bug it's fixing.

`StacktaleAppender` doesn't only warn at runtime. It validates its own XML-supplied values in `start()`, which Joran calls during configuration:

```java
zoneId = zone == null || zone.isBlank() ? ZoneId.systemDefault() : ZoneId.of(zone);
} catch (DateTimeException e) {
    addWarn("invalid zone '" + zone + "', falling back to system default", e);
```

`invalid redactPattern '…' ignored` is the same story. Those are configuration problems — a bad value in the XML, swallowed with a fallback — and their origin is the appender, so `!(s.getOrigin() instanceof StacktaleAppender)` drops them.

I ran it rather than reasoning about it. Change `<zone>UTC</zone>` to `<zone>Not/AZone</zone>` in `configuresEveryPropertyFromXmlAndReports`.

On this branch:

```
Tests run: 5, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

On main:

```
java.lang.AssertionError: Joran could not bind part of the XML — a setter name or type is wrong:
[WARN in ...StacktaleAppender[STACKTALE] - invalid zone 'Not/AZone', falling back to system
default java.time.zone.ZoneRulesException: Unknown time-zone ID: Not/AZone]
```

That failure is the whole issue in one line. Catching it is right — an unusable `<zone>` in the XML should fail this test. The sentence attached to it is wrong, because it sends you off to check setter names that are fine. Filtering by origin fixes the sentence by deleting the catch.

The discriminator isn't *who* raised the status, it's *when*. Everything this test is about — Joran's binding complaints, a rejected zone, a pattern that won't compile — lands while `doConfigure` is running. The operational warnings #205 anticipates, a write that fails or a rotation that can't proceed, only appear once events start flowing. So:

- in `configure()`, once `doConfigure` returns, record `ctx.getStatusManager().getCopyOfStatusList().size()`;
- have `assertNoJoranComplaints` assert over that prefix only, with no origin filter;
- and say what's true in the message: something warned while configuring, here it is — without naming a cause.

`invalid zone` fails again, a write failure mid-test doesn't, and nothing has to guess at origins.

If that's more machinery than you want, #205's second option is still open and I'd take it: keep catching everything, and change only the message. It fixes what was reported and removes no coverage.

Two smaller things.

`appenderWarningIsNotMistakenForJoranComplaint` calls `appender.addWarn(...)` directly. That exercises the filter, but it can't fail for the reason that would actually break it. The pipeline reaches `addWarn` through an anonymous `ReportPipeline.Host` declared inside `StacktaleAppender`, so the enclosing instance is the origin today — pull that `Host` out into its own `ContextAwareBase` some day and the origin changes, `instanceof StacktaleAppender` stops matching, and this test stays green while the filter has silently stopped working. Driving a real warning through the pipeline, the unwritable `<file>` path from the issue, pins that chain too.

And `assertThatThrownBy` is imported from `AssertionsForClassTypes`, while the file's other AssertJ import is `org.assertj.core.api.Assertions`. Same method either way; worth keeping the file on one entry point.

Happy to look again whenever you've pushed.

GabrielBBaldez:
One more instance, sharper than the `zone` one, that turned up while I was checking something unrelated: a `<file>` the writer cannot open produces

```
WARN in ...StacktaleAppender[STACKTALE] - invalid report file '…', stacktale disabled
```

Origin `StacktaleAppender`, raised during configuration. That is stacktale switched off entirely by a bad path in the XML, and it lands in the set the filter drops. Same conclusion as above — just a better argument for cutting on *when* rather than *who*.

GabrielBBaldez (approved):
This is better than what I suggested. The prefix is the right cut, and driving the operational warning through the real pipeline — configure against a directory that exists, delete it, then log — pins `append()` → `process` → `Host.warn` instead of today's shape of it. That was the part I flagged as untestable as written, and you turned it into the opposite.

I checked both cases from the review rather than reading the diff.

`<zone>Not/AZone</zone>` in `configuresEveryPropertyFromXmlAndReports`, which passed on the previous revision:

```
Tests run: 5, Failures: 1
Logback warned or errored while configuring:
[WARN in ...StacktaleAppender[STACKTALE] - invalid zone 'Not/AZone', falling back to system default]
```

Caught, and the message no longer blames a setter name.

The `<file>` case from my follow-up — a path the writer cannot open, which switches stacktale off entirely:

```
Tests run: 6, Failures: 1
Logback warned or errored while configuring:
[WARN in ...StacktaleAppender[STACKTALE] - invalid report file '…', stacktale disabled]
```

Also caught. And the suite is green on Windows, which I ran deliberately: `Files.delete` on a directory behaves differently there and the new test leans on it.

Thanks for staying with this one. The second revision is the version worth having.

PR #231, opened 2026-09-05
infra: guard the documented MCP tool surface

### LINKED ISSUE #219: infra: nothing checks the README's MCP tool list against the server
The README's MCP section opens with a count — "**Nine tools**, all read-only" — and follows it
with a list. Nothing checks either against the server.

It has already gone stale once and I did it: #218 added `repro_for` and left the README saying
"**Six tools**". It was caught by hand two PRs later, while adding two more.

That is the same failure mode as #210 and #203: a claim in a document about code that the code
does not know exists. Both of those now have a guard; this one does not.

## What to build

`scripts/check-mcp-tools.sh`, in the shape of `check-module-names.sh` and
`check-config-parity.sh`, wired into `ci.yml` beside them:

- every `tools.add(tool("<name>"` in `StacktaleMcpServer.java` has its name somewhere in
  `README.md`, `docs/mcp-setup.md` and `plugins/stacktale/README.md`;
- the count in the README's "**N tools**" line equals the number registered.

Both halves matter for different reasons. The count is what a reader trusts at a glance and the
easiest thing to leave behind; the per-name check is what tells a contributor which of the three
documents they forgot — the tool table in the plugin README is the one that gets missed, because
it is furthest from the code.

Spelling out the number in words ("Nine") is a small annoyance for a shell script. Either map
the digits, or take the opportunity to make the sentence carry the digit.

Not that it passes on a clean tree — a grep that matches nothing passes too. Add a throwaway
tool registration, confirm the script names it as undocumented, and remove it. Then take one
existing tool out of `plugins/stacktale/README.md` and confirm it names that file specifically
rather than just failing.

drexthealpha:
Closes #219

**Problem.** Three documents describe `StacktaleMcpServer`'s registrations and none of them knows when a tool is added. #218 left the README saying Six; it was caught by hand two PRs later.

**Change.** `scripts/check-mcp-tools.sh`, in the shape of `check-module-names.sh` and `check-config-parity.sh`, wired into `ci.yml` beside them. Two halves, as the issue asks:

- every `tools.add(tool("<name>"` in the server has its name in `README.md`, `docs/mcp-setup.md` and `plugins/stacktale/README.md`;
- the `**N tools**` count equals the number registered.

It found one on the first run — `plugins/stacktale/README.md` said **Six** while listing ten. Fixed here, so the guard is green.

The words are mapped rather than parsed. An unmapped count fails loudly and says to add it, so growing past the list cannot pass quietly.

**Verify.** Not just that it passes on a clean tree:

```
bash scripts/check-mcp-tools.sh
# MCP tool surface agrees with all three documents (10 tools)
```

Adding a throwaway registration names it in all three, and the count too:

```
check-mcp-tools: the server registers 'throwaway_probe', absent from README.md
check-mcp-tools: the server registers 'throwaway_probe', absent from docs/mcp-setup.md
check-mcp-tools: the server registers 'throwaway_probe', absent from plugins/stacktale/README.md
check-mcp-tools: README.md says '**Ten tools**'; the server registers 11 (Eleven)
```

Deleting one row from the plugin README names that file specifically:

```
check-mcp-tools: the server registers 'tests_covering', absent from plugins/stacktale/README.md
```

Three files, +4/-1 plus the script. No behaviour change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,6 +39,9 @@ jobs:
       - name: Config parity guard
         if: matrix.java == 21
         run: bash scripts/check-config-parity.sh
+      - name: MCP tool surface guard
+        if: matrix.java == 21
+        run: bash scripts/check-mcp-tools.sh
       # After the build, because it reads the jars' manifests rather than the poms alone —
       # the point is what actually ships, not what the pom meant to say.
       - name: JPMS module name guard
--- a/plugins/stacktale/README.md
+++ b/plugins/stacktale/README.md
@@ -16,7 +16,7 @@ Maven Central on first run.
 
 ## What you get
 
-**Six tools** over the report file, read-only, no network:
+**Ten tools** over the report file, read-only, no network:
 
 | Tool | |
 |---|---|
--- a/scripts/check-mcp-tools.sh
+++ b/scripts/check-mcp-tools.sh
@@ -1,0 +1,79 @@
+#!/usr/bin/env bash
+# Guard the documented MCP tool surface.
+#
+# StacktaleMcpServer registers the tools; three documents describe them, and none of the
+# three knows when a tool is added. #218 added `repro_for` and left the README saying "Six
+# tools"; it was caught by hand two PRs later. That is the same failure mode #210 and #203
+# already have a guard for.
+#
+# Two halves, because they fail differently:
+#
+#   every registered name appears in all three documents  -- tells you which file you forgot
+#   the "**N tools**" count equals the number registered  -- what a reader trusts at a glance
+#
+# The count is spelled in words. Rather than teach a shell script English, the words it has
+# met are mapped below; an unmapped one is reported rather than skipped, so growing past the
+# list fails loudly instead of silently passing.
+set -euo pipefail
+
+cd "$(dirname "$0")/.."
+
+server="stacktale-mcp/src/main/java/io/github/gabrielbbaldez/stacktale/mcp/StacktaleMcpServer.java"
+docs=("README.md" "docs/mcp-setup.md" "plugins/stacktale/README.md")
+
+fail=0
+
+# `|| true`: a grep that matches nothing exits 1, which `set -e` would take as this script
+# failing rather than as the finding it is.
+names="$(grep -oE 'tools\.add\(tool\("[a-z_]+"' "$server" | sed 's/.*"\(.*\)"/\1/' || true)"
+if [ -z "$names" ]; then
+  echo "check-mcp-tools: no tool registrations found in $server — has the call shape changed?" >&2
+  exit 1
+fi
+count="$(printf '%s\n' "$names" | wc -l | tr -d ' ')"
+
+for doc in "${docs[@]}"; do
+  if [ ! -f "$doc" ]; then
+    echo "check-mcp-tools: $doc is missing" >&2
+    fail=1
+    continue
+  fi
+  while IFS= read -r name; do
+    if ! grep -qF "$name" "$doc"; then
+      echo "check-mcp-tools: the server registers '$name', absent from $doc" >&2
+      fail=1
+    fi
+  done <<< "$names"
+done
+
+# The words this sentence has needed so far. Add to the map when the surface grows.
+word_for() {
+  case "$1" in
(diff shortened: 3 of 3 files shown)


GabrielBBaldez (changes requested):
This is the guard #219 wanted, and turning up the plugin README saying **Six** on the very first run is the best argument for it there could be. `word_for` reporting an unmapped number instead of skipping past it is the right instinct too — silently passing is the failure mode this whole family of scripts exists to prevent.

One hole, and it is live today rather than hypothetical.

`grep -qF "$name"` matches anywhere in the file, so a tool name that is a prefix of another registered name is satisfied by its longer sibling. `errors_since` and `errors_since_last_check` are both registered right now.

I took `errors_since` out of `README.md` entirely — the only occurrence left in the file is the one inside `errors_since_last_check`:

```
$ grep -c errors_since README.md
1
$ grep -n errors_since README.md
551:- **`errors_since_last_check`** — the loop primitive: what's 🆕 new or 🔁 still occurring
$ bash scripts/check-mcp-tools.sh
MCP tool surface agrees with all three documents (10 tools)
```

Undocument a tool and the guard reports that all three documents agree.

The same probe against `culprit_source`, which nothing else contains, behaves exactly as designed:

```
check-mcp-tools: the server registers 'culprit_source', absent from README.md
```

So the script is right and the matcher is loose by one character class.

All three documents already wrap every name in backticks, so anchoring on those is enough. I checked all ten against all three before suggesting it — ten for ten, three for three:

```bash
grep -qF "\`$name\`" "$doc"
```

A word-boundary regex works as well, if you would rather not lean on the backticks.

Two things I would leave as they are, but worth knowing you are choosing them rather than missing them.

The check runs server → documents only, so a row describing a tool that no longer exists survives as long as the count was updated too. The count catches the common case, and that direction has never bitten this repo.

`head -1` on the count phrase has the same shape: a second `**N tools**` further down a file would not be looked at. Each of the three has exactly one today, so nothing is hiding behind it. `check-plugin-versions.sh` ended up checking every occurrence after the same thought, if you want the symmetry.

Fix the matcher and this goes in.

drexthealpha:
Fixed. You were right that it was live rather than hypothetical — I reproduced your probe before changing anything and the guard did report agreement.

The matcher now anchors on the backticks:

```bash
grep -qF -- "\`$name\`" "$doc"
```

Escaped, because an unescaped backtick inside double quotes is command substitution — the first version of this fix ran every tool name as a command and then matched the empty string against every document, which passes unconditionally. Worse than the bug it replaced, so it is worth saying out loud.

Ten for ten across all three documents, as you found.

Your probe, now:

```
$ grep -c errors_since README.md
2
$ bash scripts/check-mcp-tools.sh
check-mcp-tools: the server registers 'errors_since', absent from README.md
$ echo $?
1
```

`culprit_source` still behaves the same way it did, so the change did not trade one blind spot for another:

```
check-mcp-tools: the server registers 'culprit_source', absent from README.md
```

I also took the `head -1` symmetry with `check-plugin-versions.sh`, since you pointed at it — the count now reads every `**N tools**` in a file rather than the first:

```
$ printf '\n**Six tools**\n' >> README.md
$ bash scripts/check-mcp-tools.sh
check-mcp-tools: README.md says '**Six tools**'; the server registers 10 (Ten)
```

Clean tree still passes: `MCP tool surface agrees with all three documents (10 tools)`.

The other one you flagged — documents → server, so a row for a deleted tool survives if the count moved with it — I have left alone deliberately, as you suggested. It is a different check with a different failure mode and it belongs in its own change rather than smuggled into this one.

GabrielBBaldez (approved):
Both closed, and I re-ran the probes rather than reading them.

`errors_since` taken out of `README.md`, its only remaining occurrence the one inside `errors_since_last_check`:

```
check-mcp-tools: the server registers 'errors_since', absent from README.md
```

A second, stale count appended further down the same file:

```
check-mcp-tools: README.md says '**Six tools**'; the server registers 10 (Ten)
```

`culprit_source` behaves as it did before, and a clean tree still passes.

The comment explaining why the backticks are the anchor is worth as much as the fix — the next person through will know it is load-bearing rather than decorative.

Thanks for this one. The guard earned itself on the first run by catching the plugin README, and now it holds for the one pair of names that would have slipped past it.

PR #232, opened 2026-09-06
docs: add deployment guidance

### LINKED ISSUE #60: Deployment story: containers, stdout-only shops, and multiple instances
﻿The review found the container/K8s story is the biggest structural adoption gap. \`errors-ai.log\` is a file on local disk; a 12-factor / stdout-only shop has nowhere for it to live, and multiple pod replicas each write their own. \`emitReportsToLogger\` (one report = one event on \`stacktale.reports\`) is the intended container answer but is not positioned as such.

Add a Deployment section: (1) the stdout/shipper path via \`emitReportsToLogger\` + how a collector picks it up; (2) file path guidance for a writable volume; (3) multi-instance behavior; (4) whether \`errors-ai.log\` should be git-ignored (yes) with a snippet.

---

## Guidance for a first-time contributor

This is documentation-only â€” no Java, no build. Everything you need is already in the repo;
the work is arranging it into one section a reader can follow.

**Where it goes:** a new `## Deployment` section in `README.md`, after *Configuration* and
before *Guarantees*. Match the surrounding voice: short paragraphs, concrete snippets, no
marketing.

**What to write, in order:**

1. **Containers / stdout-only shops.** The answer is `emitReportsToLogger` â€” it emits each
   report as a *single* log event on the `stacktale.reports` logger, so a collector picks it
   up whole instead of 40 interleaved lines. It is documented under *Configuration*; here,
   frame it as "you have no writable disk, use this". Show the Spring property and the
   Logback snippet, and mention that a shipper should treat that logger as one message.
2. **Writable volume.** If you *can* mount a volume, `file` should point at it. Say plainly
   that the container's ephemeral filesystem loses reports on restart.
3. **Multiple replicas.** Each instance writes its own file â€” that is by design, and
   `SECURITY.md` already states the single-writer assumption. Say what a reader should do:
   one file per instance, or the stdout path above. Do **not** invent a shared-file mode.
4. **Git-ignore it.** `errors-ai.log` is a build artifact, not source. Give the one-line
   `.gitignore` snippet, including the rotated backups (`errors-ai.log*`).

**Ground every claim in the code or existing docs** â€” if you are unsure whether something is
true, ask in the issue rather than guessing. A wrong deployment instruction is worse than a
missing one.

**Verify:** render the README locally (or in the PR preview) and check the links and code
fences. No build required â€” say so in the PR.

janithcd:
## Summary

* add deployment guidance for containers and stdout-only environments
* document persistent-volume configuration and multi-replica file handling
* clarify that generated report files and rotated backups should be excluded from version control

## Details

* shows the Spring Boot and Logback configuration for `emitReportsToLogger`
* explains that each report is emitted as one event through the `stacktale.reports` logger
* recommends placing report files on a persistent writable volume
* documents the single-writer requirement and one-file-per-instance approach

## Verification

* `git diff --check`
* reviewed the section order, relative link, and fenced configuration examples
* no build run because this is a documentation-only change

Closes #60


GabrielBBaldez (changes requested):
Four sections in the order the issue asked for, grounded in what the repo already says, and the volume, replica and gitignore parts are all correct. Thank you for picking it up.

The problem is in the first section, and it started with me — the issue told you to frame `emitReportsToLogger` as "you have no writable disk, use this". That is not what the code does, and you had no reason to doubt it.

`emitReportsToLogger` mirrors the file, it does not replace it. `ReportPipeline.process` opens with:

```java
if (writer == null || parked || SELF_LOGGER.equals(event.loggerName())
        || REPORTS_LOGGER.equals(event.loggerName())) return;
```

`writer` is null whenever the destination could not be opened — `isActive()` is literally `writer != null`. And the emission sits downstream of a successful write:

```java
writer.append(rendered);
...
if (settings.emitReportsToLogger()) host.emitReport(rendered);
```

I ran it rather than trusting the read. An appender with `emitReportsToLogger=true` and a `<file>` under a path that cannot be created, one error logged, a `ListAppender` attached to `stacktale.reports`:

```
PROBE events on stacktale.reports: 0
WARN in ...StacktaleAppender[STACKTALE] - invalid report file '.../blocker/errors-ai.log',
     stacktale disabled
```

Zero events, and the appender says it out loud: disabled. So a reader on a read-only root filesystem follows this section, sets the flag, and gets nothing at all — which is worse than not having the section.

What it should say instead: stacktale always needs a writable path, and `emitReportsToLogger` puts a copy of every report on stdout in addition to it. If the root filesystem is read-only, point `file` at whatever writable tmpfs the platform does give you — `/tmp`, an `emptyDir` mount — and the emission works from there. That keeps the section aimed at the same reader and makes it true.

One more sentence worth adding while you are in there, because it will catch someone out: only full reports are emitted. Recurrence summaries go to the file alone —

```java
case SUMMARY -> {
    writer.append(renderer.renderSummary(fingerprint, decision.count(), decision.lastSeenMillis()));
```

— so a stdout-only reader sees each first occurrence and never the "seen 47 times" line that follows it.

The rest stands as written. I will fix the issue text too, so the next person does not inherit it.

janithcd:
Thanks for the clarification, @GabrielBBaldez ! Updated in f25a18a.

The section now explicitly requires a writable report file, includes `/tmp/errors-ai.log` in both configuration examples, and explains that logger emission depends on successful file writing. It also clarifies that only full reports are emitted, while recurrence summaries remain file-only.

The other three sections are unchanged, and `git diff --check` passes. Ready for another review.

GabrielBBaldez (changes requested):
Both corrections are in and accurate — leading with the writable file, the `/tmp` and `emptyDir` guidance, and saying plainly that a file which cannot be opened disables stacktale.

One phrase left, and it is mine: I wrote "seen 47 times" in the review and that is not what the follow-up line is called. `seen:` is a field *inside* a report block. The follow-up is the repeat line, which this same README already shows two hundred lines above your section:

```
━ #c73cf755 repeated 47× ━
```

Swap `seen N times` for `repeated N×` and this is done.

Issue #234, opened 2026-09-06
emitReportsToLogger drops recurrence summaries: stdout sees the first occurrence and never the count


GabrielBBaldez:
With `emitReportsToLogger=true`, a full report reaches the `stacktale.reports` logger and a recurrence summary does not. In `ReportPipeline.process`:

```java
case REPORT -> {
    ...
    writer.append(rendered);
    ...
    if (settings.emitReportsToLogger()) host.emitReport(rendered);
}
case SUMMARY -> {
    writer.append(renderer.renderSummary(fingerprint, decision.count(), decision.lastSeenMillis()));
    deduper.confirmWritten(fingerprint, decision.count());
    summariesWritten.incrementAndGet();
}
```

The `SUMMARY` arm never calls `host.emitReport`.

For anyone reading the file this is invisible — both land there. For the reader this flag exists for, the one whose reports only ever appear on stdout, it means the same error looks like it happened once. The summary is where "and then it happened another 46 times" lives, so it is arguably the line that matters most to somebody watching a shipper.

## What to change

Emit the summary the same way the report is emitted. It is a small change in the `SUMMARY` arm, and the ordering matters for the same reason it does above: `deduper.confirmWritten` must stay after the append, because a failed append has to leave the count pending for `close()`'s `drainPending()`. The emission goes last, after the state is settled — a failing shipper must not undo dedup state.

Then say it in the README. The Configuration section describes `emitReportsToLogger` as one report per event; once summaries go too, that sentence needs to cover both.

## How to check you got it right

- A test that logs the same error twice inside the dedup window with `emitReportsToLogger=true`, and asserts two events on `stacktale.reports` rather than one. `LogbackXmlConfigTest` shows how to attach a listener to a named logger; `DeduperTest` shows how to cross a window without sleeping.
- Assert on the *content* of the second event, not just the count — a summary that arrives with the report's own text would pass a count-only assertion and be useless.
- `mvn verify` green, and no golden file touched: this changes what reaches the logger, not what is written to the report file.

Say the word here and it's yours.

Found while reviewing #232. See also #233, which is about the larger question of whether this flag should work without a file at all.

vinkdc:
Hi! I'm happy to help out with this as my first open source contribution. I can add the summary logger emission, regression coverage that checks the content of the second event, and update the README as described. Please assign it to me. Thank you!

GabrielBBaldez:
Being fixed in #237.

PR #237, opened 2026-09-07
fix(core): emit recurrence summary to external logger (#234)

vinkdc:
The SUMMARY branch in ReportPipeline.process() was writing dedup summaries to the report file but not forwarding them to the external logger (stacktale.reports), unlike the BLOCK branch. This caused downstream aggregators to miss recurrence events.

- Extract rendered summary into a local variable
- Call host.emitReport(rendered) after confirmWritten, mirroring BLOCK
- Add integration test: two identical errors -> assert second logger event contains 'repeated 2x' and is not a full block
- Update README emitReportsToLogger description

## What & why

Closes #

## How it was verified

- [ ] `mvn verify` is green (JDK 17+)
- [ ] Documentation-only — nothing to build

## Checklist

- [ ] The issue was claimed with a comment before starting (see
      [CONTRIBUTING](https://github.com/stacktale/stacktale/blob/main/CONTRIBUTING.md#claim-the-issue-before-you-start)
      — didn't claim it? Open the PR anyway, just say so)
- [ ] One logical change (unrelated fixes belong in their own PR)
- [ ] Behavior changes arrive with the test that demanded them; bug fixes start from a
      failing test
- [ ] **No golden-file test changed** — `st/1` is a public API. If a golden test did
      change, that is a deliberate format change: say so below, and update
      [docs/FORMAT.md](https://github.com/stacktale/stacktale/blob/main/docs/FORMAT.md)
- [ ] New config property? It has a setter (Joran naming), a default that preserves
      current behavior, coverage in `LogbackXmlConfigTest`, and a row in the README
      config table


GabrielBBaldez (changes requested):
The placement is the part that was easy to get wrong, and you got it right without being told: after `writer.append` and after `confirmWritten`, so a shipper that throws cannot undo dedup state. That is the contract the REPORT arm spells out two lines above, and this arm now honours it.

I ran it rather than reading it. Your test passes, and removing just the emission line makes it fail for the right reason:

```
Expected size: 2 but was: 1
```

One thing to settle first, and it is the README line rather than the code.

`renderSummary` is called from two places. The one you changed runs while events are flowing; the other is in `close()`, draining whatever the dedup window still had pending when the app shut down. Five identical errors, `emitReportsToLogger=true`:

```
file:    ━ #4359adb0 repeated 2× (last 16:44:34.646) ━
         ━ #4359adb0 repeated 5× (last 16:44:34.650) ━

logger:  ━━━ ERROR #4359adb0 ━━━ …
         ━ #4359adb0 repeated 2× (last 16:44:34.646) ━
```

The final count — the one that says this error happened five times, not twice — is the one that stays behind. So `Also emit each report block and recurrence summary as ONE event` promises more than the code delivers, which is the shape of the problem #234 was about to begin with.

Either way out is fine by me:

- narrow the README sentence to the summaries emitted during the run, or
- emit at the drain site as well. I measured that rather than assuming it — adding the call there took the probe from 2 events to 3, so a Logback context on its way down is not in the way. A JVM shutdown hook is a harder case and I have not measured that one.

The second is still one logical change — summaries reach the logger — so it belongs in this PR rather than a follow-up, if you want it.

Not yours, and mine to deal with: the `storm:` line has the same shape. Written to the file, never emitted, and it is the line that says reports are being dropped.

One small thing: the template's `Closes #` is still blank, so #234 will not close itself when this merges.

GabrielBBaldez:
Followed through on the storm line rather than leaving it as a remark: #238, measured the same way. It touches the arm right above yours, so it waits on this one.

PR #247, opened 2026-09-16
fix(mcp): a bounded search must not answer as if it had finished

### LINKED ISSUE #244: tests_covering says "none" on any repo with more than 500 test files — and tells the agent that means write a test
`Workspace.walk` caps what it collects:

```java
paths.filter(Files::isRegularFile)
        .filter(p -> !isSkipped(p))
        .filter(accept)
        .limit(500)
```

The cap is applied *after* the filter, so `testSources()` returns the first 500 test files the walk happens to reach and stops. Nothing says it stopped.

That would be harmless if the answer were hedged. It is the opposite — the negative is the answer `tests_covering` presents as decisive:

> none: no test source names PaymentService.charge.
>
> Searched 500 file(s) under src/test. … nothing naming it is a strong signal that the failing path is untested. `repro_for` gives you the call and its arguments to write one from.

So on a repo with more than 500 test files the tool can tell an agent, with confidence and a next action attached, to go write a reproduction test that already exists.

## Measured

A temp tree with 600 filler test files plus one that does name `PaymentService.charge`:

```
PROBE test files on disk: 601
PROBE answer: none: no test source names PaymentService.charge.
PROBE says-none: true
PROBE Searched 500 file(s) under src/test. …
```

`Searched 500 file(s)` reads as *this repo has 500 test files*, not *I gave up at 500*.

## Why it has never bitten here

stacktale has **52** test sources. The tool is being developed against a tree an order of magnitude under its own cap, and 500 test files is unremarkable in the Java repos this is aimed at.

## What to change

The truncation has to stop being invisible. In rough order of how much I like them:

- **Don't cap the scan, cap the output.** The walk is cheap next to `Files.readString` on each hit; what needs bounding is how many hits get printed, and that is already handled separately (`matched > 20`, `MAX_LINES_PER_FILE`).
- If a bound on the scan stays, it must reach the answer: `none` becomes something like *no match in the first 500 of N test files*, and it must not carry the "strong signal, go write one" language, because at that point it is not a strong signal.

`filesNamed`, behind `culprit_source`, shares the same `walk` and the same cap. Its failure is milder — "No file named X under …" instead of a wrong file — but it is the same sentence: an answer stated as fact that was really a give-up.

Build a tree with more test files than the cap, with the only match past it, and assert the answer is not `none`. The probe above is the whole test; `@TempDir` and a loop writing filler files is all it needs.

Assert on the wording too, not just on the match count — the bug here was never that the search was bounded, it was that the sentence did not say so.

GabrielBBaldez:
Closes #244.

`Workspace.walk` collected 500 matching files and stopped. Nothing carried that upward, so `tests_covering` answered:

```
none: no test source names PaymentService.charge.

Searched 500 file(s) under src/test. … nothing naming it is a strong signal that the
failing path is untested. repro_for gives you the call and its arguments to write one from.
```

having read the first 500 of however many there were. That negative is the answer the tool is read for — it is the one that sends an agent off to write a reproduction test — so above the cap it could send it to write one that already exists. And `Searched 500 file(s)` reads as a count of the repository, not as a give-up.

stacktale has **52** test sources, an order of magnitude under its own cap. That is why it was never seen here.

## Change

A walk now returns whether it ran out, and both callers have to answer for it.

- The bound is 20 000 rather than 500. Walking directory entries is cheap next to `readString`; what needs limiting is how much of the answer gets printed, and `MAX_LINES_PER_FILE` plus the 20-file cap already do that.
- A truncated search answers `inconclusive` and **loses** the sentence that made the negative actionable.
- An exhaustive one keeps it, and now says `Searched all N file(s)`.
- `culprit_source` gets the same treatment: it used to say `No file named X` when what had happened was that the walk failed.

A test seam takes the bound as a constructor argument — writing 20 000 files to disk to reach a branch is not a test.

## Verified

Four new tests in `WorkspaceTest`, and they fail against the old behaviour. Silencing just the truncation flag:

```
WorkspaceTest.aTruncatedSearchDoesNotClaimNothingNamesTheMethod:36
  a search that gave up must not answer as if it had finished:
  none: no test source names PaymentService.charge.

WorkspaceTest.culpritSourceSaysTheSearchFailedRatherThanThatTheFileIsAbsent:73

Tests run: 5, Failures: 2
```

With the fix: `Tests run: 41, Failures: 0` across the module, and `check-mcp-tools.sh` green.

One thing worth knowing rather than glossing: `filesNamed` applies its bound *after* the name filter, so for `culprit_source` the bound is effectively unreachable and the branch is there for the failed-walk case. The test reaches it that way — a root that is not there — rather than pretending the bound did it.


GabrielBBaldez:
Pushed one more: the count was not the only quiet bound.

`Files.walk(root, MAX_DEPTH)` stops at twelve levels and says nothing, and `services/payments/src/test/java/com/acme/x/y/Z.java` is already eleven — so a module-per-directory repository reaches it without being unusual, and a test source below it produced the same confident `none`. A directory sitting exactly at the limit is the evidence there was more underneath, so that now counts as running out too.

Same discipline on the test: silence just the depth detection and it fails with

```
the walk stopped at its depth limit and still answered as if it had seen the whole tree:
none: no test source names PaymentService.charge.
```

`Tests run: 42, Failures: 0` for the module.

